PR #20, opened 2022-09-23
Conditional rendering to show results

krischarbonneau:
### Description

This PR adds conditional rendering to the home page so that when a user enters valid information, the form disappears and a response message is shown indicating either the status of the application or that there was a problem retrieving the information (in this case when a name is entered that has no info). I've also changed the error handling for the ESRF field so it catches when the number is less than 8 digits. I've also removed the borders on the response messages as they aren't inline with the form anymore. Lastly, I added a button that is shown below the response that resets all form data via a handler and updated the focus styling on the `ActionButton` component.

1. Pull in branch
2. Type `npm run dev`
3. Enter valid and invalid data in the form to ensure it works as intended (ie. enter 1234 into the ESRF field for the 8 digit error to pop up, enter John for given name and then Jack to see success message vs problem retrieving)

### DIFF
--- a/pages/home.js
+++ b/pages/home.js
@@ -25,9 +25,24 @@ export default function Home(props) {
     }
   }
 
+  const handleReset = async (e) => {
+    e.preventDefault()
+    //clear form data, errors & results
+    setErrorSummary()
+    setEsrfError()
+    setGivenNameError()
+    setSurnameError()
+    setBirthDateError()
+    setEsrf()
+    setGivenName()
+    setSurname()
+    setBirthDate()
+    setResponse()
+  }
+
   const handleSubmit = async (e) => {
     e.preventDefault()
-    //clear erros & results
+    //clear errors & results
     setErrorSummary()
     setEsrfError()
     setGivenNameError()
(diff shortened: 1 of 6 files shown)


pages/home.js
```diff
@@ -92,79 +107,87 @@ export default function Home(props) {
 
   return (
     <>
-      <h1>{props.content.header}</h1>
-      <p>{props.content.description}</p>
+      <h1 className="mb-4">{props.content.header}</h1>
+      {!response ? <p>{props.content.description}</p> : undefined}
```
s-laugh, line 111: We could pop this p tag down into the same condition on line 120

pages/home.js
```diff
@@ -92,79 +107,87 @@ export default function Home(props) {
...
-          type="date"
-        />
-        <ActionButton
-          type="submit"
-          text={props.content.checkStatus}
-          style="primary"
-        />
-      </form>
+      {response ? (
+        <div>
```
s-laugh, line 182: Again, we could move this up, put it under the same condition use to show the response

s-laugh (commented):
Just a thought on moving some things around so we only use the `!response` condition once.

pages/home.js
krischarbonneau: fair points on both suggestions, changed!

s-laugh (approved):
Great work!

PR #24, opened 2022-09-28
Update dependencies to latest version

sebastien-comeau:
### Description

List of proposed changes:

- Update all dependencies to latest version

Testing tools have been updated and everything seems to work as attended.

### Additional Notes

React 18 have a lot of improvments and it doesn't support old browsers like IE which is good. It adds new hooks like `useId` and `useDeferredValue` which I've used in other projects.
https://blog.appsignal.com/2022/04/13/whats-new-in-react-18.html

**Important!!!**
You need to run `npm install` to update your local `node_modules`


krischarbonneau (approved):
Everything works as intended, thanks!

PR #25, opened 2022-09-28
Keep form filled on unsuccessful request

krischarbonneau:
### Description

This PR adds a condition to the reset handler that checks `typeof response`. If `object` is returned, the form data is reset when the user presses the "Go back" button, but if anything else is returned then when the user selects the back button the form fields will remain filled so they can check for errors/typos. This works because the only time `object` will ever be returned is if the request to the API was successful. The other scenarios will return `string`, which are the cases where the user is shown the text "We are unable to determine the status of your application at this time using this service".

1. Pull in branch
2. Ensure you have the environment variable for the API set in your `.env.local` (msg me for this if you don't)
3. Type `npm run dev`
4. Enter the following test data 
```
ESRF: 35934S87
First Name: Clara
Last Name: Rénard
DOB: 1982-12-05
```
5. Pressing submit with the above data should yield a response object with `status: completed`; meaning that if you select the back button, the form will be cleared
6. Now enter the below test data
```
ESRF: 12345678
First Name: Steve
Last Name: Smith
DOB: 1994-12-02
```
7. Pressing submit should show the message `We are unable to determine the status of your application at this time using this service`. Pressing the back button should keep the form fields filled because `response` is equal to either `not-found` or `non-unique`

### Additional Notes

If for whatever reason we move away from conditional rendering in the future, we could also store the form values in localstorage and populate the form on page load based on a callback


pages/home.tsx
```diff
@@ -53,6 +54,8 @@ const Home: FC = () => {
   >()
   const [errorSummary, setErrorSummary] = useState<ErrorSummaryItem[]>([])
 
+  console.log(typeof response)
```
Maxim-Lam, line 57: Might want to remove the console.log but all good apart from that

Maxim-Lam (changes requested):
Apart from the console log, everything is good!

Maxim-Lam (approved):
LGTM

PR #26, opened 2022-09-28
imrpove the error handling on the home page form. reduce complexity s…

Maxim-Lam:
### Description

List of proposed changes:
Improved error handling on the form. If/when new conditions will be added to each field, it will be easier to check for them without changing too much code.

test form functionality unchanged


krischarbonneau (approved):
Works well, good job!

PR #27, opened 2022-09-29
Add react-query library and api hooks to fetch data

sebastien-comeau:
### Description

List of proposed changes:

- Add TanStack React Query library
- Add `useCheckStatus` hook to handle check status API calls with the use of react-query hook
- Refactor `/api/check-status` endpoint
- Map API 3 passport statuses in english and french
- Refactor `home` page with the use of `useCheckStatus` hook

If `PASSPORT_STATUS_API_BASE_URI` environment variable is undefined then `/api/check-status` endpoint will fetch the data from mock passport statuses data stored in `__mocks__\passportStatusesMock.js` else should fetch from the API.

### Additional Notes

React Query library

> React Query is often described as the missing data-fetching library for React, but in more technical terms, it makes fetching, caching, synchronizing and updating server state in your React applications a breeze.

https://tanstack.com/query/v4/docs/overview


krischarbonneau (approved):
Meaty PR! This is fantastic, works as expected. Thanks!

PR #30, opened 2022-10-03
Add next translate

sebastien-comeau:
### Description

List of proposed changes:

- Add next-translate library is to keep the translations as simple as possible in a Next.js environment.

1. Run `npm install` to install the new package
2. Run the application in `dev` or `production`
3. Browse the application and toogle the language
4. Fields error message should be translated as well

### Additional Notes

next-translate information and guide
https://github.com/aralroca/next-translate


components/Footer.tsx
```diff
@@ -43,12 +43,11 @@ const Footer: FC<FooterProps> = ({
   footerNav1,
   footerNav2,
 }) => {
-  const commonLocale = useCommonLocale()
-
+  const { t } = useTranslation('common')
```
s-laugh, line 46: Components shouldn't be responsible for their own translation; all the text should get passed to it's props from the page level. We may want to revisit this after, as it would be significant change to all the components edited in this PR.

krischarbonneau (approved):
Similar implementation to Service Canada Labs, looks good to me!

PR #35, opened 2022-10-03
Added landing page, updated ActionButton component

krischarbonneau:
### Description

This PR adds a simple landing page that either sends the user to our form if they have an ESRF or to the existing intake form if they don't have an ESRF (this may change in the future/we may add another form w/o ESRF but until the data feed solution is finalized this way is considered MVP).

I updated the `ActionButton` component to render a link using `Next/Link` if supplied an `href`.

There were a couple typos in the locale files that were resulting in some console warnings that I fixed as well.

1. Pull in branch
2. Type `npm run dev`
3. Upon selecting your language at the splash screen, you should be redirected to the new landing page.
4. If you select the left button (no ESRF) you will be redirected to the existing intake form
5. If you select the right button (you have your ESRF) you will be redirected to our form

### Additional Notes

The landing page in its current state is very simple, so I'm open to ideas/suggestions on sprucing it up

### DIFF
--- a/__tests__/pages/status.test.js
+++ b/__tests__/pages/status.test.js
@@ -1,0 +1,36 @@
+import { render, screen } from '@testing-library/react'
+import '@testing-library/jest-dom'
+import Status from '../../pages/status'
+import { useRouter } from 'next/router'
+import { useCheckStatus } from '../../hooks/api/useCheckStatus'
+
+// mocks useRouter to be able to use component' router.asPath
+jest.mock('next/router', () => ({
+  useRouter: jest.fn(),
+}))
+jest.mock('../../components/InputField', () => () => {
+  return <mock-modal data-testid="test-modal" />
+})
+jest.mock('../../components/ActionButton', () => () => {
+  return <mock-modal data-testid="button-modal" />
+})
+jest.mock('../../components/ErrorSummary', () => () => {
+  return <mock-modal data-testid="errors-modal" />
+})
+jest.mock('../../hooks/api/useCheckStatus')
+
+describe('Check status page page', () => {
+  beforeEach(() => {
+    useRouter.mockImplementation(() => ({
+      pathname: '/',
+      asPath: '/',
+    }))
+    useCheckStatus.mockImplementation(() => ({}))
+  })
+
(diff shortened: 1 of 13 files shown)


Maxim-Lam (approved):
Nice Kris, everything looks good. As for the look of the landing page, it doesn't matter to me seeing as the functionality is there and works great. Maybe styling could be improved but that's just nitpicking. 

Looks good

components/ActionButton.tsx
```diff
@@ -28,7 +31,11 @@ const ActionButton: FC<ActionButtonProps> = ({
       break
   }
 
-  return (
+  return href ? (
+    <Link href={href}>
+      <a className={classStyle}>{text}</a>
+    </Link>
+  ) : (
```
s-laugh, line 38: This seems like such a different thing, that I think it would be best to build as a separate component all together; or just use the React/Link.

pages/home.tsx
```diff
@@ -1,222 +1,34 @@
-import InputField from '../components/InputField'
```
s-laugh, line 1: The re-use of this page name `home` for something else, will break the history. I think it's better to add the splash page under another name; then later we can come back to renaming files.

s-laugh (changes requested):

components/ActionButton.tsx
krischarbonneau: This is the way it was done in the design system in order to create a "link button". I think there is merit in this being here since we can re-use the styling but we can have a discussion about it

s-laugh: hmmm right, I might take a crack at separating the use while keeping the commonality; after you merge.

__tests__/pages/status.test.js
```diff
@@ -19,7 +19,7 @@ jest.mock('../../components/ErrorSummary', () => () => {
 })
 jest.mock('../../hooks/api/useCheckStatus')
 
-describe('Home page', () => {
+describe('Check status page page', () => {
```
s-laugh, line 22:
```suggestion
describe('Check status page', () => {
```

s-laugh (approved):
Nice!

PR #37, opened 2022-10-04
helper updates

s-laugh:
### Description

List of proposed changes:

- change dependabot branch separator from `/` to `_`
- change dast test to a schedule so we don't get failed notifications because the job was run too recently

Really only testable after merge by perceiving the changes happen as expected.


krischarbonneau (approved):
Looks good!

PR #43, opened 2022-10-04
Middleware to add /en to URL

krischarbonneau:
### Description

I've added a middleware that intercepts URL requests with the `default` locale (a dummy locale that was created since Next ignores adding the `defaultLocale` to the URL) and appends `/en` in front of the path.

This is the current official way of doing this in Next, which you can read more about [here](https://nextjs.org/docs/advanced-features/i18n-routing#prefixing-the-default-locale).

I've also updated the metadata title to what Steph mentioned. Since the `Index` page is supposed to be bilingual, I added a condition to the middleware that intercepts requests to `/en` or `/fr` and sends the user back to `/`. I also changed the metadata to be bilingual.

1. Pull in branch
2. Type `npm run dev`
3. Upon starting the app you will be brought to `http://localhost:3000/`, ensure that when you add `en` or `fr` to the end of that links that you are redirected back to `/`
4. Click `English` or `Francais`, you should see the proper language in the path for both the landing and status pages


middleware.ts
```diff
@@ -0,0 +1,26 @@
...
+  if (
+    req.nextUrl.pathname.startsWith('/_next') ||
+    req.nextUrl.pathname.includes('/api/') ||
+    PUBLIC_FILE.test(req.nextUrl.pathname)
+  ) {
+    return
+  }
+
+  if (req.nextUrl.locale === 'default' && !req.nextUrl.pathname.endsWith('/')) {
+    return NextResponse.redirect(new URL(`/en${req.nextUrl.pathname}`, req.url))
```
s-laugh, line 16: In this case we don't have a language, maybe we want to send the back to the index so that it can be selected, rather than defaulting?? (This however does maintain current function to default to en)

pages/index.tsx
```diff
@@ -1,19 +1,17 @@
 import Image from 'next/image'
 import Link from 'next/link'
 import MetaData from '../components/MetaData'
-import useTranslation from 'next-translate/useTranslation'
 
 const Index = () => {
-  const { t } = useTranslation('common')
```
s-laugh, line 7: I thinking it might be a good idea to keep this so ALL the text stays in one place; add a common.json to the "default" language to pull the text?

s-laugh (commented):
Just a few thoughts/suggestions

middleware.ts
s-laugh: Suggesting that we never default to a language; so if one isn't specified we re-route to the index where the user is forced to make a selection.

Maybe I miss understood how the routing works; if on `/en/landing`, could the code just specify `/status` (link) would the router get me to `/en/status` without this?

pages/index.tsx
krischarbonneau: Next-i18next actually ignores the default locale folder entirely so this wouldn't work. You'd see `meta.title` as the text for the title for example. This wasn't always the case but it was a feature they implemented to work with how Next handles prefixing default locales.

See more [here](https://github.com/i18next/next-i18next/commit/f9a653966cd52fbb50bcbbd545369dbad50b489b)

That said, I do need to delete the default folder under locales since it literally does nothing.

s-laugh: ahh sucky, how would you feel about a direct import of a .json file (obj) into here just for those values?

krischarbonneau: That would work, sure.

middleware.ts
s-laugh: hmm okay; I think what I'm after isn't worth the nitpickyness; so I'll drop it :)

PR #46, opened 2022-10-04
cleaning up some things with status page

s-laugh:
### Description

List of proposed changes:

- Enable a Reset to bring you back to landing on success getting a status
- Convert page to be static
- Remove some un-used code
- fix the cron

If you get a status response, the "go back" should be "restart" and bring you back to the landing page to start the flow again.

### DIFF
--- a/pages/status.tsx
+++ b/pages/status.tsx
@@ -15,10 +15,6 @@ import StatusInfo from '../components/StatusInfo'
 import { GetStaticProps } from 'next'
 import Router from 'next/router'
 
-export const getStaticProps: GetStaticProps = async () => ({
-  props: {},
-})
-
 const Status: FC = () => {
   const { t } = useTranslation('status')
 
@@ -183,4 +179,9 @@ const Status: FC = () => {
     </Layout>
   )
 }
+
+export const getStaticProps: GetStaticProps = async () => ({
+  props: {},
+})
+
 export default Status
--- a/pages/status.tsx
+++ b/pages/status.tsx
@@ -6,12 +6,18 @@ import ErrorSummary, {
   getErrorSummaryItem,
 } from '../components/ErrorSummary'
 import Layout from '../components/Layout'
-import { CheckStatusReponse, CheckStatusRequestBody } from './api/check-status'
+import { CheckStatusRequestBody } from './api/check-status'
 import { useCheckStatus } from '../hooks/api/useCheckStatus'
 import { useFormik } from 'formik'
 import * as Yup from 'yup'
 import useTranslation from 'next-translate/useTranslation'
 import StatusInfo from '../components/StatusInfo'
+import { GetStaticProps } from 'next'
+import Router from 'next/router'
+
+export const getStaticProps: GetStaticProps = async () => ({
+  props: {},
+})
 
 const Status: FC = () => {
   const { t } = useTranslation('status')
@@ -67,13 +73,13 @@ const Status: FC = () => {
       )
   }, [formik, t])
 
-  const handleReset: MouseEventHandler<HTMLButtonElement> = useCallback(
-    (e) => {
-      e.preventDefault()
-      formik.resetForm()
-    },
-    [formik]
-  )
+  // const handleReset: MouseEventHandler<HTMLButtonElement> = useCallback(
+  //   (e) => {
+  //     e.preventDefault()
+  //     formik.resetForm()
+  //   },
+  //   [formik]
+  // )
 
   const handleGoBack: MouseEventHandler<HTMLButtonElement> = useCallback(
     (e) => {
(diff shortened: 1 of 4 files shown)


pages/status.tsx
```diff
@@ -67,13 +73,13 @@ const Status: FC = () => {
...
   }, [formik, t])
 
-  const handleReset: MouseEventHandler<HTMLButtonElement> = useCallback(
-    (e) => {
-      e.preventDefault()
-      formik.resetForm()
-    },
-    [formik]
-  )
+  // const handleReset: MouseEventHandler<HTMLButtonElement> = useCallback(
```
sebastien-comeau, line 76: This commented block of code should be removed. We can retreive it in previous commits if needed.

pages/status.tsx
```diff
@@ -93,14 +99,16 @@ const Status: FC = () => {
       {}
       {checkStatusReponse !== undefined ? (
         <StatusInfo
-          handleGoBackClick={handleGoBack}
-          goBackText={t('go-back')}
+          handleGoBackClick={
+            checkStatusReponse ? () => Router.push('/landing') : handleGoBack
```
sebastien-comeau, line 103: Move this if condition inside `handleGoBack` and set `onGoBackClick` prop with `handleGoBack` as before. I think we should rename `handleGoBackClick` to `onGoBackClick` to be consistent with other event props naming convention.

pages/status.tsx
```diff
@@ -93,14 +99,16 @@ const Status: FC = () => {
       {}
```
sebastien-comeau, line 99: Remove this empty JSX curly braces

pages/status.tsx
sebastien-comeau, line 76: This commented block of code should be removed. We can retreive it in previous commits if needed.

pages/status.tsx
sebastien-comeau, line 99: Remove this empty JSX curly braces

pages/status.tsx
sebastien-comeau, line 103: I think moving this if condition inside `handleGoBack` and set `onGoBackClick` prop with `handleGoBack` as before would make more sense.

We should also rename `handleGoBackClick` to `onGoBackClick` to be consistent with other event property naming convention, as example `onClick`, `onChange`, etc...

pages/status.tsx
```diff
@@ -6,12 +6,18 @@ import ErrorSummary, {
...
+import { CheckStatusRequestBody } from './api/check-status'
 import { useCheckStatus } from '../hooks/api/useCheckStatus'
 import { useFormik } from 'formik'
 import * as Yup from 'yup'
 import useTranslation from 'next-translate/useTranslation'
 import StatusInfo from '../components/StatusInfo'
+import { GetStaticProps } from 'next'
+import Router from 'next/router'
+
+export const getStaticProps: GetStaticProps = async () => ({
```
sebastien-comeau, line 18: I suggest to move `getStaticProps` before `export default Status` as you would see in next.js examples.

sebastien-comeau (changes requested):

pages/status.tsx
sebastien-comeau: https://nextjs.org/docs/basic-features/data-fetching/get-static-props#using-getstaticprops-to-fetch-data-from-a-cms

pages/status.tsx
s-laugh: I'd rather leave it for now; we are getting requirements from BAs soon that will change some things.

pages/status.tsx
s-laugh: I honestly hate that that's how examples look; its so very backwards as this function is always called before the actual component. It's far more logical to keep it at the top.

pages/status.tsx
sebastien-comeau: Sounds good

pages/status.tsx
sebastien-comeau: I see your point but it's a convention that you will see everywhere with next.js. I like to follow next.js convention.

pages/status.tsx
sebastien-comeau: It's why I was using 2 different status component before. 1 for when you receive an actual status `StatusInfo` and one when status is null `StatusInfoUnavailable`. It made the code easier to follow.

pages/status.tsx
s-laugh: okay.... :(

pages/status.tsx
s-laugh: If your suggesting it goes back to a single condition now, I'm okay to do that.

sebastien-comeau: For now, yes.

PR #48, opened 2022-10-06
Switch to next-i18next

sebastien-comeau:
### Description

List of proposed changes:

- Replace `next-translate` with `next-i18next`

Make sure the localizations load correctly as before.

### Additional Notes

next-translate requires us to add the `/pages` files in our docker image and they are not supose to be there. It's a known issue that I faced in other project with next-translate. Someone oppened an issue in next-translate repo:
https://github.com/aralroca/next-translate/issues/892

next-i18next repo and guide
https://github.com/i18next/next-i18next


pages/status.tsx
```diff
@@ -1,19 +1,30 @@
...
-import { CheckStatusRequestBody } from './api/check-status'
-import { useCheckStatus } from '../hooks/api/useCheckStatus'
-import { useFormik } from 'formik'
-import * as Yup from 'yup'
-import useTranslation from 'next-translate/useTranslation'
 import StatusInfo from '../components/StatusInfo'
-import { GetStaticProps } from 'next'
-import Router from 'next/router'
+
+const PageWrapper: FC<{ children: JSX.Element }> = ({ children }) => {
```
s-laugh, line 19: This shouldn't be separated out into another component.

pages/status.tsx
```diff
@@ -69,14 +80,6 @@ const Status: FC = () => {
       )
   }, [formik, t])
 
-  // const handleReset: MouseEventHandler<HTMLButtonElement> = useCallback(
-  //   (e) => {
-  //     e.preventDefault()
-  //     formik.resetForm()
-  //   },
-  //   [formik]
-  // )
```
s-laugh, line 78: Please keep this in for now; we might need it with the business requirements.

s-laugh (changes requested):
All the translation changes look good.
If you want to suggest the other changes to status page, please at least put them in a separate PR.

pages/status.tsx
sebastien-comeau: Why not? Othwise I would need to duplicate the following code at 3 occasions (each conditional return)

```tsx
<Layout>
    <h1 className="mb-4">{t('header')}</h1>
    ( // conditional content here }
</Layout>
```

sebastien-comeau: I will revert it as it was before.

pages/status.tsx
```diff
@@ -87,12 +90,11 @@ const Status: FC = () => {
   )
 
   //if the api failed, fail hard to show error page
-  if (checkStatusError) throw new Error(checkStatusError.message)
+  if (checkStatusError) throw checkStatusError
```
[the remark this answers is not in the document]
sebastien-comeau: `ApiError` is a class that extends `Error`. Throwing a new Error or the ApiError directly will act the same.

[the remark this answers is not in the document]
sebastien-comeau: I've reverted my changes.

s-laugh (approved):
Looks good, works well!
Thank you for keeping this PR to just the translation. If you want to suggest the other changes made earlier in the PR, please submit another PR and we can chat about it.

PR #50, opened 2022-10-07
Add react-query-devtools to find and reduce stale queries

sebastien-comeau:
### Description

List of proposed changes:

- Add react-query-devtools to analyse react-query 
- Change the way `useCheckStatus` is used

Install the packages `npm install` then run the application `npm run dev` and you will see a react-query floating button at left corner that will open `react-query-devtools`.

You should see the initial query then when the form is submitted, another entry for that query will be saved.

Before this change it had inactive queries for everyt formik values change:
image

### Additional Notes

Devtools Documentation
https://tanstack.com/query/v4/docs/devtools

> By default, React Query Devtools are only included in bundles when process.env.NODE_ENV === 'development', so you don't need to worry about excluding them during a production build.


krischarbonneau (approved):
Clean PR, works great on dev

s-laugh (commented):
I'm a bit indifferent to this... solving a problem we don't have, but good to have, but also still ships to production even if it says it turns off - we're still effected by vulnerabilities to it.

Q: If it did get turned on in production - would it leak encrypted calls?

package.json
```diff
@@ -23,7 +23,8 @@
     "@opentelemetry/api": "^1.2.0",
     "@opentelemetry/auto-instrumentations-node": "^0.33.1",
     "@opentelemetry/sdk-node": "^0.33.0",
-    "@tanstack/react-query": "^4.9.0",
+    "@tanstack/react-query": "^4.10.1",
+    "@tanstack/react-query-devtools": "^4.10.1",
```
s-laugh, line 27: It seems like this can be put into `devDependencies` (--save-dev) and the npm build/start will still work. 
I'd be a little happier with that as node shouldn't package it at all for production (risk reduction).

s-laugh (changes requested):

package.json
sebastien-comeau: I don't think it make sense to save it as dev dependency because it is used in the application and it's not a dev tool that help to build the application.

https://tanstack.com/query/v4/docs/devtools
> By default, React Query Devtools are only included in bundles when process.env.NODE_ENV === 'development', so you don't need to worry about excluding them during a production build.

sebastien-comeau:
> I'm a bit indifferent to this... solving a problem we don't have, but good to have, but also still ships to production even if it says it turns off - we're still effected by vulnerabilities to it.
> 
> Q: If it did get turned on in production - would it leak encrypted calls?

It helped me to solve a problem actually. I changed the logic when using `useStatusCheck` hooks to not filling `react-query` client cache with keys that is not valid. More than you would type in the fields more disabled keys were registered in the client cache.

We are not caching any encrytped data that you wouldn't see while fetching data from the `API`.

package.json
s-laugh: Just because the package promises to not execute in production, doesn't mean they won't have bugs or be susceptible to attacks - for every package we include in our production build, we are also susceptible to all the security issues (despite the controls in place). As this package is designed to only be used in development; it should only be package for development. This helps to reduce our security risk for something that is unnecessarily needed in a production build.

sebastien-comeau:
I've created a new PR #59 that will not include `react-query-devtools`

PR #53, opened 2022-10-07
Update button classname and skip to content button styling

krischarbonneau:
### Description
Was pointed out to me that the `LinkButton` component was using template literals for `className`. We aren't evaluating any expressions or adding in custom styling via props so this PR switches it to be more in line with the other button component which just uses a string. During testing, I also noticed that the Skip to main content button was not accessible at all when focused, so I've gone ahead and mimicked the [Canada.ca](https://www.canada.ca) styling. I've experimented adding the dark orange inner ring on focus like Canada.ca has it, so it can also be added to the rest of our buttons if we want to follow that design pattern.

1. Pull in branch
2. Type `npm run dev`
3. Make sure button styling on the splash page shows as expected.
4. Navigate to `/en/landing`
5. Tab into the Skip to main content button and ensure that it is accessible

### Additional Notes
We aren't really utilizing the skip button currently and though there isn't really any content beyond the buttons and the form we should probably still add the `#mainContent` link on each of our pages at some point

### DIFF
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -18,7 +18,7 @@ const Header: FC = () => {
       >
         <a
           id="skipToMainContent"
-          className="bg-blue-800 text-white px-2 focus:outline-black-solid hover:bg-gray-dark"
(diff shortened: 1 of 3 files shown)


s-laugh (approved):
We can change how the styles are done; it's all a trade off, long files for clean PR changes

PR #54, opened 2022-10-07
Remove translation from components

s-laugh:
### Description

Components should be dumb (take and spit out the text they are given; they should not be deciding what text to present)

List of proposed changes:

- Remove the translation from components; move to pages, passed as props
- Cleaned up some tests

All pages still render and translate properly.


krischarbonneau (approved):
Looks and works well 👍

PR #59, opened 2022-10-11
Reduce stale queries

sebastien-comeau:
### Description

List of proposed changes:

- Reduce stale queries when filling the form

Status page works properly.

### Additional Notes

I installed `react-query-devtools` and debugged `react-query`. Before this change it had staled queries for every formik values changes.

image

Devtools Documentation
https://tanstack.com/query/v4/docs/devtools

I didn't installed `react-query-devtools` permanently because it needs to be installed as an application dependency and not dev dependency. This will make sure this package is never published in production even if the documentation said it shouldn't.

> By default, React Query Devtools are only included in bundles when process.env.NODE_ENV === 'development', so you don't need to worry about excluding them during a production build.

### DIFF
--- a/pages/status.tsx
+++ b/pages/status.tsx
@@ -1,4 +1,4 @@
-import { FC, MouseEventHandler, useCallback, useMemo, useState } from 'react'
+import { FC, MouseEventHandler, useCallback, useMemo } from 'react'
 import { GetStaticProps } from 'next'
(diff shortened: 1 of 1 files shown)


s-laugh (approved):
I see nothing wrong with this; but it did raise the fact that we'll need you to walk the team though what's actually going on with `formik`

PR #61, opened 2022-10-12
988 no match status

Maxim-Lam:
### Description

List of proposed changes:

Added a LinkSummary component that displays a list of links. 

Display LinkSummary for no match results, showing three links provided in the user story.

1. enter false but valid data.
2. Test the links bring you to the appropriate places.

### Additional Notes

@s-laugh , @krischarbonneau  Please let me know if there is a specific way this should be styled and I will alter it. I left it pretty bare bones as to make it easy to change.

Also if you guys know of a better way to store the LinkSummaryItems outside of status.tsx, pls let me know.

### DIFF
--- a/components/LinkSummary.tsx
+++ b/components/LinkSummary.tsx
@@ -7,14 +7,15 @@ export interface LinkSummaryItem {
 }
 
 export interface LinkSummaryProps {
-  id: string
+  title?: string
   links: LinkSummaryItem[]
 }
 
-const LinkSummary: FC<LinkSummaryProps> = ({ id, links }) => {
+const LinkSummary: FC<LinkSummaryProps> = ({ title, links }) => {
   return (
-    <section id={id}>
-      <ul className="list-none pb-6 ml-4">
+    <section className="mt-5">
+      <p>{title}</p>
+      <ul className="list-disc pb-6 ml-4">
         {links.map(({ href, text }, index) => (
           <li key={index} className="text-link-default">
             <Link href={href}>{text}</Link>
--- a/components/LinkSummary.tsx
+++ b/components/LinkSummary.tsx
@@ -1,0 +1,28 @@
+import { FC } from 'react'
+import Link from 'next/link'
+
+export interface LinkSummaryItem {
+  href: string
+  text: string
+}
+
+export interface LinkSummaryProps {
+  id: string
+  links: LinkSummaryItem[]
+}
+
+const LinkSummary: FC<LinkSummaryProps> = ({ id, links }) => {
+  return (
+    <section id={id}>
+      <ul className="list-none pb-6 ml-4">
+        {links.map(({ href, text }, index) => (
+          <li key={index} className="text-link-default">
+            <Link href={href}>{text}</Link>
+          </li>
+        ))}
+      </ul>
+    </section>
+  )
+}
+
+export default LinkSummary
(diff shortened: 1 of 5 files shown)


pages/status.tsx
```diff
@@ -26,6 +27,21 @@ const initialValues: CheckStatusRequestBody = {
...
+    },
+    {
+      href: 'https://www.canada.ca/en/immigration-refugees-citizenship/services/canadian-passports/contact-passport-program.html',
+      text: t('no-match-status-links.contact-program'),
+    },
+    {
+      href: 'https://eservices.canada.ca/en/reservation/',
+      text: t('no-match-status-links.appointment'),
+    },
+  ]
```
s-laugh, line 43: I think you could have the whole object come from the translation. The links need to be translated too (french links for the french page)

pages/status.tsx
```diff
@@ -118,14 +134,20 @@ const Status: FC = () => {
...
-          goBackStyle="primary"
-          checkAgainText={t('check-again')}
-        >
-          <p className=" mb-6 text-2xl">{t('unable-to-find-status')}</p>
-        </StatusInfo>
+        <>
+          <LinkSummary
+            id={'noMatchLinkSummary'}
+            links={noMatchHref}
+          ></LinkSummary>
```
s-laugh, line 141: I think this should come after the status Info, as that would be the primary information the user is looking for.

components/LinkSummary.tsx
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export interface LinkSummaryProps {
+  id: string
+  links: LinkSummaryItem[]
+}
+
+const LinkSummary: FC<LinkSummaryProps> = ({ id, links }) => {
+  return (
+    <section id={id}>
```
s-laugh, line 16: I'm not sure the id is adding any value to this component...

components/LinkSummary.tsx
```diff
@@ -0,0 +1,28 @@
...
+      <ul className="list-none pb-6 ml-4">
```
s-laugh, line 17: I think this should be styled like a normal list with bullets; and would be good to include a `p` tag with some text above, for now can give it the text "Other way to get your status:" (translated)

s-laugh (commented):
This looks great! Answering your questions with a few suggestions.

components/LinkSummary.tsx
Maxim-Lam: My original thought process was maybe for some form of accessibility purpose or incase we needed to get the entire section by id for something but if it's not necessary I will remove it

pages/status.tsx
```diff
@@ -26,6 +27,15 @@ const initialValues: CheckStatusRequestBody = {
...
   const { t } = useTranslation('status')
 
+  const getLinkSummaryItems: any = () => {
+    const LSI: any = t('no-match-status-links', { returnObjects: true })
+    const noMatchHref: LinkSummaryItem[] = []
+    for (const keys in LSI) {
+      noMatchHref.push(LSI[keys] as LinkSummaryItem)
+    }
+    return noMatchHref
+  }
```
s-laugh, line 37: I don't think all this should be necessary, would just need to included `t('no-match-status-links', { returnObjects: true })` as the prop for the links...

s-laugh (approved):
Awesome!

PR #65, opened 2022-10-13
Add the new email esrf api endpoint

s-laugh:
### Description

List of proposed changes:

- Creating the API and Hook (Light weight and just the basics for now, expect more to be added when call from client is added)
- Remove unused mock file

API should get responses with env variable set.


Maxim-Lam (approved):
LGTM, clean!

sebastien-comeau:
Sorry, forgot to submit my review after I added comments.

PR #66, opened 2022-10-14
Fix ternary operators should not be nested

sebastien-comeau:
### Description

List of proposed changes:

- Fix sonarlint warning : Fix ternary operators should not be nested

image

Status page should act like before.

### Additional Notes

Ternary operators should not be nested rule information
https://rules.sonarsource.com/javascript/RSPEC-3358

### DIFF
--- a/pages/status.tsx
+++ b/pages/status.tsx
@@ -107,125 +107,139 @@ const Status: FC = () => {
       footer={t('common:footer', { returnObjects: true })}
     >
       <h1 className="mb-4">{t('header')}</h1>
-      {checkStatusReponse ? (
-        <StatusInfo
-          handleGoBackClick={() => Router.push('/landing')}
-          goBackText={t('reset')}
-          goBackStyle="primary"
-          checkAgainText={t('check-again')}
-        >
-          <p className="mb-6 text-2xl">
-            {`${t('status-is')} `}
-            <strong>
-              {t(`status.${checkStatusReponse.status}`, {
-                defaultValue: checkStatusReponse.status,
-              })}
-            </strong>
-            .
-          </p>
-        </StatusInfo>
-      ) : checkStatusReponse === null ? (
-        <>
-          <StatusInfo
-            handleGoBackClick={handleGoBack}
-            goBackText={t('go-back')}
-            goBackStyle="primary"
-            checkAgainText={t('check-again')}
-          >
-            <p className=" mb-6 text-2xl">{t('unable-to-find-status')}</p>
-          </StatusInfo>
-          <LinkSummary
-            title={t('no-match-title')}
-            links={lsItems}
-          ></LinkSummary>
-        </>
-      ) : (
-        <form onSubmit={formik.handleSubmit} id="form-get-status">
-          <p>{t('description')}</p>
-          {errorSummary.length > 0 && (
-            <ErrorSummary
-              id="error-summary-get-status"
-              summary={t('common:found-errors', {
-                count: errorSummary.length,
-              })}
-              errors={errorSummary}
+      {(() => {
+        if (checkStatusReponse) {
+          return (
+            <StatusInfo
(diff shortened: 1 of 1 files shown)


s-laugh:
Well... I do disagree with this... 

Clean and easy reading of code requires a function to have a single input and a single output.
In addition, braking a if into multiple if statements doesn't help one bit, (i'd expect each if to execute, but the return inside a if prevents that from happening); it should be using if else so show a single flow of logic (if it wasn't for the fact that we are looking for undefined and null, a switch would be the better option).

sebastien-comeau:
> Well... I do disagree with this...
> 
> Clean and easy reading of code requires a function to have a single input and a single output. In addition, braking a if into multiple if statements doesn't help one bit, (i'd expect each if to execute, but the return inside a if prevents that from happening); it should be using if else so show a single flow of logic (if it wasn't for the fact that we are looking for undefined and null, a switch would be the better option).
> 
> _I do understand that sometime readability can just be a preference thing; IMO this change does not help._

It's a kind of conditional rendering your will see often.

image

https://reactjs.org/docs/conditional-rendering.html

Or an element variable or inline If with logical && operator can be used.

s-laugh:
> It's a kind of conditional rendering your will see often.

I know... it pains me. 

In general I give myself one exception to the single in/out rule when writting, that's to fail fast out of a function. 
So more like what we see here on the page (it could be a return if warranted, in this case throw is more direct at achieving the result we want) https://github.com/DTS-STN/passport-status/blob/6daaefe609c638fc97c52010e7f5f4b69ab77df5/pages/status.tsx#L94

I do agree that complex if conditions should not use the short operators `? : &&` as it becomes hard to read, instead should write out the long format if. But I wouldn't go out of my way to change what's already been put in place unless working on it directly.

If @Maxim-Lam and @krischarbonneau want this change, I won't object; it's just not my prefrence.

Maxim-Lam:
> > It's a kind of conditional rendering your will see often.
> 
> I know... it pains me.
> 
> In general I give myself one exception to the single in/out rule when writting, that's to fail fast out of a function. So more like what we see here on the page (it could be a return if warranted, in this case throw is more direct at achieving the result we want)
> 
> https://github.com/DTS-STN/passport-status/blob/6daaefe609c638fc97c52010e7f5f4b69ab77df5/pages/status.tsx#L94
> 
> I do agree that complex if conditions should not use the short operators `? : &&` as it becomes hard to read, instead should write out the long format if. But I wouldn't go out of my way to change what's already been put in place unless working on it directly.
> 
> If @Maxim-Lam and @krischarbonneau want this change, I won't object; it's just not my prefrence.

I can see both sides of the argument being valid, If statements are just more easily readable at a glance and it's scope is much more visible. No objections either way

krischarbonneau:
> > It's a kind of conditional rendering your will see often.
> 
> I know... it pains me.
> 
> In general I give myself one exception to the single in/out rule when writting, that's to fail fast out of a function. So more like what we see here on the page (it could be a return if warranted, in this case throw is more direct at achieving the result we want)
> 
> https://github.com/DTS-STN/passport-status/blob/6daaefe609c638fc97c52010e7f5f4b69ab77df5/pages/status.tsx#L94
> 
> I do agree that complex if conditions should not use the short operators `? : &&` as it becomes hard to read, instead should write out the long format if. But I wouldn't go out of my way to change what's already been put in place unless working on it directly.
> 
> If @Maxim-Lam and @krischarbonneau want this change, I won't object; it's just not my prefrence.

I think this is where I stand too. Generally I don't like nesting ternary operators since it reduces readability but in this case I don't think it was too complex. I'm okay with switching to the long form if conditions to improve readability and if for whatever reason we add more logic to this page it will help in the long run as well.

krischarbonneau (approved):
I'm okay with this if the others are.

PR #67, opened 2022-10-14
Added cancel button with modal dialog

krischarbonneau:
### Description
This PR adds the `CancelModal` component which renders an `ActionButton` where, upon click, the user is presented with a modal dialog. I've made the component re-usable in the sense that we can supply the modal dialog as many buttons as we want (so perhaps renaming the component to just `Modal` would be better? Open to thoughts on this). For our use case, when a user clicks the cancel button they are asked `Are you sure you want to go back?`. If they answer `Yes`, they will be redirected to the landing page. If they select `No`, they dialog will close and they can continue to fill out the form.

This also brings in the `react-focus-on` package, which locks focus within the modal, removes scrolling functionality (and hides the scrollbar) when focus is within the modal, and also applies aria-hidden to the rest of the page when within the modal to prevent screen readers from announcing anything on the page aside from the modal content.

1. Pull in branch
2. Type `npm run dev`
3. Navigate to `/status`
4. Click Cancel
5. Clicking `Yes` should redirect you back to `/landing`
6. Clicking `No` should close the modal

### DIFF
--- a/components/CancelModal.tsx
+++ b/components/CancelModal.tsx
@@ -1,0 +1,56 @@
+import { FC, MouseEventHandler } from 'react'
+import ActionButton from './ActionButton'
+import { ActionButtonProps } from './ActionButton'
+import { FocusOn } from 'react-focus-on'
+
+export interface CancelModalProps {
+  buttonText: string
+  description: string
+  isOpen: boolean
+  onClick: MouseEventHandler<HTMLButtonElement>
+  buttons: ActionButtonProps[]
+}
+
+const CancelModal: FC<CancelModalProps> = ({
+  buttonText,
+  description,
+  buttons,
+  onClick,
+  isOpen,
+}) => {
+  return (
+    <>
+      <ActionButton text={buttonText} onClick={onClick} style="primary" />
+      {isOpen ? (
+        <FocusOn autoFocus={false}>
+          <div
+            className="fixed top-0 left-0 w-screen h-full flex justify-center items-center"
+            style={{ background: 'rgba(71, 71, 71, 0.8)' }}
+          >
+            <div className="p-4 bg-white border-2 border-black">
+              <p className="font-body">{description}</p>
+              <div className="flex space-x-2 mx-4">
+                {buttons.map((button) => {
+                  return (
+                    <div key={button.text}>
+                      <ActionButton
+                        text={button.text}
+                        onClick={button.onClick}
+                        style={button.style}
+                        type={button.type}
+                      />
+                    </div>
+                  )
+                })}
+              </div>
+            </div>
+          </div>
+        </FocusOn>
+      ) : (
+        ''
+      )}
+    </>
+  )
+}
+
+export default CancelModal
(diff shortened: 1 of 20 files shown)


pages/status.tsx
```diff
@@ -179,12 +181,38 @@ const Status: FC = () => {
...
             required
           />
-          <ActionButton
-            disabled={isCheckStatusLoading}
-            type="submit"
-            text={t('check-status')}
-            style="primary"
-          />
+          <div className="flex flex-wrap">
+            <div id="form-get-status" className="py-1 pr-2">
```
s-laugh, line 185: I'm surprised moving the id fixed the cypress test... I don't think this id should have been moved; it should be an id for the whole form, not just the submit button.

components/CancelModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+const CancelModal: FC<CancelModalProps> = ({
+  buttonText,
+  description,
+  buttons,
+  onClick,
+  isOpen,
+}) => {
+  return (
+    <>
+      <ActionButton text={buttonText} onClick={onClick} style="primary" />
```
s-laugh, line 23:
```suggestion
      <ActionButton text={buttonText} onClick={onClick} />
```
The style needs to be "default", we can't have two primary buttons on the same page.
You could consider passing the style to this component... but still needs to be set to default.
_FYI, this rule doesn't apply to the buttons in the modal, as we typical avoid showing bias towards an option_

s-laugh (changes requested):
Looks good! Quick fix on the button style. I do agree making it a generic modal sounds like a good idea; but leave that decision to you (can be done later when needed).

components/CancelModal.tsx
krischarbonneau: Good catch. We'll need to update the default styling to be in line with the Canada.ca design pattern in a separate PR. It's mostly fine, just need to update the colors slightly.

components/CancelModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+      {isOpen ? (
```
sebastien-comeau: Use inline If with Logical && Operator.
https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator

components/CancelModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+        <FocusOn autoFocus={false}>
+          <div
+            className="fixed top-0 left-0 w-screen h-full flex justify-center items-center"
+            style={{ background: 'rgba(71, 71, 71, 0.8)' }}
+          >
+            <div className="p-4 bg-white border-2 border-black">
+              <p className="font-body">{description}</p>
+              <div className="flex space-x-2 mx-4">
+                {buttons.map((button) => {
```
sebastien-comeau: You should spread the ActionButtonProps like this:

```jsx
{buttons.map((buttonProps) => (
                  <ActionButton key={buttonProps.text} {...buttonProps} />
                ))}
```

krischarbonneau: Much cleaner, thanks.

pages/status.tsx
s-laugh: ahh right, maybe instead, the cy.get should target an id specifically on the button?

krischarbonneau: Yeah, that would be the best solution.

s-laugh (approved):
Great Work

PR #68, opened 2022-10-14
Add reusable error messaging

s-laugh:
### Description

List of proposed changes:

- Use some i18next nesting so that repeated text is only in our translation file in one spot.

All error message still show properly and translate.


krischarbonneau (approved):
Works well!

PR #69, opened 2022-10-14
add the env for api url to docker

s-laugh:
### Description

List of proposed changes:

- Add env variable to docker at runtime

Validate that it's using data from the API and not our mock file

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,5 +27,7 @@ COPY --from=build /build/tracing.js ./
 RUN VERSION_NEXT=`node -p -e "require('./package.json').dependencies.next"`&& npm install --no-package-lock --no-save next@"$VERSION_NEXT"
 
 # Runtime envs -- will default to build args if no env values are specified at docker run
+ARG PASSPORT_STATUS_API_BASE_URI
+ENV PASSPORT_STATUS_API_BASE_URI=$PASSPORT_STATUS_API_BASE_URI
 
 CMD npm run start


sebastien-comeau:
I don't thing you need to bake this variable when building the container image because it's being read by the server only. You could push the environment virable when running the container and it should work.

gregory-j-baker:
I agree with @sebastien-comeau on this one. By using a build-time argument, you're (potentially<sup>††</sup>) committing to building environment-specific images, which should really be avoided.

I feel like build args should be limited to things that are build-scoped.. for example `BUILD_TIMESTAMP` or `BUILD_COMMIT`.

For _server side_ variables, pass them in at runtime. For _client side_ variables, consider passing them in at runtime, then fetching them via an API call... which I demoed [in this project](https://github.com/DTS-STN/nextjs-config-poc).

<sup>††</sup>when client side variables are involved

s-laugh:
This is pretty standard way of handling these for us; we'd have to do a lot of work to back track on how we manage our builds to change this; unless you have a simple solution. At this point it's not worth that level of effort.

s-laugh:
I also forgot to mention that I believe it doesn't completely bake it in, as it's added just before the `start` command, they can be injected and overridden on `docker run`.

Although, to be clear, that is where our challenge arises of adding them at runtime; this build doesn't specifically control the docker run command; being able to extract a variable from vault in that command.

gregory-j-baker:
You are correct that the environment variable is being set inside the container before the `start` command.. but adding it as a build statement also hardcodes it into the container image.

``` sh
[gbaker@blackwater docker]$ cat Dockerfile
from bash
ARG MSG
ENV MSG=${MSG}
CMD echo $MSG

[gbaker@blackwater docker]$ docker build --build-arg MSG=Hello . --tag test
[ ...build junk... ]

[gbaker@blackwater docker]$ docker run test # Note: baked in value
Hello

[gbaker@blackwater docker]$ docker run --env MSG=World test
World
```

I'm not sure I understand the comment about pulling secrets from the vault. If you do have a secret in the vault and you specify it at build time, **you are baking your secret into your container image**, which is a big problem.

I think it's okay to build in some variables as sane defaults, but you have to be really selective over which ones. You don't want (for example) to hardcode a prod URL into your container image and then accidentally hit prod when you forget to override that variable at runtime.

**References:**
- <https://docs.docker.com/engine/reference/builder/#env>

gregory-j-baker:
Agreed in a side chat to revisit environment variables down the road, so don't block this merge because of me.

sebastien-comeau (approved):
There was an agreement in a side chat to revisit environment variables down the road.

PR #71, opened 2022-10-17
Added Link summary component to match result screen

Maxim-Lam:
### Description

List of proposed changes:

added LinkSummary component to match result screen. changing of key name in json file.

enter valid information for application.

(
35934S87,
Clara,
Renard,
1982-12-08
)

view results screen link summary and test hrefs for each link.

### DIFF
--- a/public/locales/en/status.json
+++ b/public/locales/en/status.json
@@ -45,7 +45,8 @@
     "no-button": "No",
     "description": "Are you sure you want to go back?"
   },
-  "other-ways-to-status": "Other ways to get your status.",
+  "other-ways-to-status": "Contact us",
+  "no-match-title": "Other ways to get your status.",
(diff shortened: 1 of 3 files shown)


public/locales/en/status.json
```diff
@@ -45,7 +45,7 @@
     "no-button": "No",
     "description": "Are you sure you want to go back?"
   },
-  "no-match-title": "Other ways to get your status.",
+  "other-ways-to-status": "Other ways to get your status.",
```
s-laugh, line 48: You are still using `no-match-title` so don't delete it. The match title might need to be just "Contact us" as they already got their status, they  don't need to get it again.

Maxim-Lam: you're right!

s-laugh (approved):
Great!

PR #73, opened 2022-10-17
updating the api and hook a little bit

s-laugh:
Keeping the API really light - It's really just a security pass through, no need to add any additional logic.

Addressees some of the issues mentioned after merge on #65


krischarbonneau (approved):
Looks good to me

PR #74, opened 2022-10-17
moving some files around for consistency

s-laugh:
### Description

List of proposed changes:

- Mainly eliminating additional folders (could be added back when project complexity increases)


pages/api/check-status.ts
```diff
@@ -94,7 +70,7 @@ export const fetchPassportStatusAPI = async (
   )
 
   if (response.ok) {
-    const passportStatusesResponse = (await response.json()) as PassportStatuses
+    const passportStatusesResponse = await response.json() //) as CheckStatusReponse[]
```
sebastien-comeau, line 73: Returned JSON from API should be mapped with TS Type otherwise it will be type of `any` which is discouraged. If the API change the JSON structure then you can't easily validate this change in your code. If you have a type that match the API then updated it will ensure your code reflect the changes.

lib/CheckStatusHook.ts
```diff
@@ -3,14 +3,13 @@ import { ApiError } from 'next/dist/server/api-utils'
 import {
```
sebastien-comeau, line 3: Filename should be `useCheckStatus`.

pages/api/check-status.ts
s-laugh: In complex calls where there is parsing of the data and not everything gets passed on I can completely understand. But the objects are nearly identical, short of being nested in the return object. This API is already doing way more than it should; all it should be is a (redundant) pass though.

lib/CheckStatusHook.ts
s-laugh: File names don't have hard fast rules, adding the key word "use" provides no context as to what this file is for. This updated name make that far more clear.

sebastien-comeau: `useSomething` is well known in the react to be a hook. Component naming convention is PascalCase. Functions, variables and props should be in camelCase. TS types, interfaces and enums should be in PascalCase. I suggest to follow what's is being used within react community.

example from MUI for hook naming:
https://github.com/mui/material-ui/tree/master/packages/mui-material/src/useMediaQuery

pages/api/check-status.ts
s-laugh: 1. You've convinced me... 
2. Over zealous ESDC IT Security... (a client browser should not have rights to call an "internal" api - even if in this case that API is open - it's firmly against our architecture standards)

s-laugh: Added the Types back.

lib/CheckStatusHook.ts
s-laugh: The file isn't just that function; It's all the functions related to the hook.
My personal preference would also have me change the name, but I am leaving it.

lib/StatusTypes.ts
```diff
@@ -0,0 +1,34 @@
...
+  status: string
+}
+
+/**
+ * Map passport status object returned by the API to the client
+ * filter external API fields
+ * @param param0 Passport status object returned by the API
+ * @returns Frontend check status object
+ */
+export function MapToCheckStatusReponse({
```
sebastien-comeau, line 28: Function name should be in camelCase.

lib/CheckStatusHook.ts
sebastien-comeau: It's either that we move the fetcher function in his own file or we get ridd of it and move the method logic inside useQuery `queryFn` parameter.

s-laugh: I don't see the need of that in this moment and would be out of scope for this PR.

lib/StatusTypes.ts
s-laugh: We don't have to follow every global convention. Infact most places suggest a team internally decide and agree on the standards they want to follow; to my knowledge we don't have one.

sebastien-comeau: It's a javascript convention in general. I'm always trying to follow the language naming convention. To my knowledge, in react you use PascalCase for component names other than that it's camelCase.

sebastien-comeau: I feel it's important to follow the industry conventions of the language you are working with. We can have a discussion outside a PR.

sebastien-comeau (approved):

PR #79, opened 2022-10-18
Fix property 'id' is missing in test

sebastien-comeau:
### Description

List of proposed changes:

- Fix property 'id' missing in `StatusInfo.test.tsx`

Execute the npm script `tsc:validate` complile and validate TS.

### Additional Notes

I would be a good idea to run this script in the build pipeline.


s-laugh (approved):
`tsc:validate` could be run as part of the linting...

PR #81, opened 2022-10-18
Add next-i18next config appendNamespaceToMissingKey

sebastien-comeau:
### Description

List of proposed changes:

- Configure next-i18next appendNamespaceToMissingKey to true

If you call `useTranslation("status")` with a default namespace `status` then there is no need to append the namespace with the key when using `t` function like `t("go-back")`. If the key is not found then only the key is displayed. Configuring appendNamespaceToMissingKey to true will append the namespace to the missing key.

Before:
image

After:
image

### Additional Notes
https://www.i18next.com/overview/configuration-options#missing-keys

### DIFF
--- a/next-i18next.config.js
+++ b/next-i18next.config.js
@@ -1,6 +1,12 @@
+// @ts-check
+
+/**
+ * @type {import('next-i18next').UserConfig}
+ **/
 module.exports = {
   i18n: {
     locales: ['default', 'en', 'fr'],
     defaultLocale: 'default',
   },
+  appendNamespaceToMissingKey: true,
 }


next-i18next.config.js
```diff
@@ -1,6 +1,12 @@
+// @ts-check
```
s-laugh, line 1: Not sure about a TypeScript check on a JavaScript file...

next-i18next.config.js
```diff
@@ -1,6 +1,12 @@
...
+
+/**
+ * @type {import('next-i18next').UserConfig}
```
s-laugh, line 4: Is this commented out, or is it doing something; would like to understand what it's doing if it is.

next-i18next.config.js
sebastien-comeau: Information about @ts-check
https://www.typescriptlang.org/docs/handbook/intro-to-js-ts.html#ts-check

PR #83, opened 2022-10-19
Change check-status API endpoint to GET method

sebastien-comeau:
### Description

List of proposed changes:

- Change the API `/api/check-status` endpoint method to GET and use query instead of body

Run the application, fill the form, validate the result and watch the requests with DevTools.

### Additional Notes
URLSearchParams.get()
https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/get


sebastien-comeau:
I'm trying to fix tests.

lib/CheckStatusHook.ts
```diff
@@ -1,23 +1,24 @@
...
-  const response = await fetch('/api/check-status', {
-    ...init,
-    method: 'POST',
-    body: JSON.stringify(request),
-  })
+  const query = Object.keys(checkStatusRequest)
+    .map(
+      (key) => `${key}=${checkStatusRequest[key as keyof CheckStatusRequest]}`
+    )
+    .join('&')
```
s-laugh, line 13: Why not build this into a common function, and re-used it in the api and the test?

pages/api/check-status.ts
```diff
@@ -81,24 +81,36 @@ export default async function handler(
   req: NextApiRequest,
   res: NextApiResponse<CheckStatusReponse | string>
 ) {
-  if (req.method !== 'POST') {
-    return res.status(405).send(`Invalid request method ${req.method}`)
+  if (req.method !== 'GET') {
+    res.status(405).send(`Invalid request method ${req.method}`)
+    return
```
s-laugh, line 86: Why break this into two lines?

pages/api/check-status.ts
```diff
@@ -81,24 +81,36 @@ export default async function handler(
...
   }
 
-  const body = JSON.parse(req.body) as CheckStatusRequestBody
+  const { searchParams } = new URL(req.url ?? '', `http://${req.headers.host}`)
+  const checkStatusRequest: CheckStatusRequest = {
+    birthDate: searchParams.get('birthDate') ?? '',
+    esrf: searchParams.get('esrf') ?? '',
+    givenName: searchParams.get('givenName') ?? '',
+    surname: searchParams.get('surname') ?? '',
+  }
```
s-laugh, line 95: This is what feels "janky" and error prone to a simple typo and why I didn't want to pursue this change atm.

pages/api/check-status.ts
```diff
@@ -81,24 +81,36 @@ export default async function handler(
...
+        )
+      : await fetchPassportStatusMOCK(checkStatusRequest)
+    res.status(200).json(response)
   } catch (error) {
-    logger.error(error)
     if ((error as Error).constructor.name === 'ApiError') {
       const apiError = error as ApiError
-      return res.status(apiError.statusCode).send(apiError.message)
+      res.status(apiError.statusCode).send(apiError.message)
+      return
```
s-laugh, line 110: Again, why split this on two lines?

pages/api/check-status.ts
```diff
@@ -81,24 +81,36 @@ export default async function handler(
...
     }
+
+    logger.error(error)
```
s-laugh, line 113: We want to log all the errors, not just the ones that make it this far; this should be moved to the top of the catch.

__tests__/api/check-status.test.ts
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * NextApiRequest and node-mocks-http createRequest types union
+ * @see: https://github.com/howardabrams/node-mocks-http/issues/255#issuecomment-1136674043
+ */
+export type ApiRequest = NextApiRequest & ReturnType<typeof createRequest>
+
+/**
+ * NextApiResponse and node-mocks-http createResponse types union
+ * @see: https://github.com/howardabrams/node-mocks-http/issues/255#issuecomment-1136674043
+ */
```
s-laugh, line 15: could remove the duplicate comment

__tests__/api/check-status.test.ts
```diff
@@ -0,0 +1,52 @@
+import handler from '../../pages/api/check-status'
+import { createMocks, createRequest, createResponse } from 'node-mocks-http'
+import { CheckStatusRequest } from '../../lib/StatusTypes'
+import { NextApiRequest, NextApiResponse } from 'next'
+
+/**
+ * NextApiRequest and node-mocks-http createRequest types union
+ * @see: https://github.com/howardabrams/node-mocks-http/issues/255#issuecomment-1136674043
+ */
+export type ApiRequest = NextApiRequest & ReturnType<typeof createRequest>
```
s-laugh, line 10: don't need to export this, should be local to the test (at least while in this file)

lib/CheckStatusHook.ts
sebastien-comeau: I thought about this too and was on my list on thing to improve. Searching online to simple way of doing it and found an simpler solution with the usage of `URLSearchParams`.

```js
const obj = {
  page: 2,
  limit: 10,
  filter: 'js',
};

const result = '?' + new URLSearchParams(obj).toString();
console.log(result); // 👉️ "?page=2&limit=10&filter=js"
```

pages/api/check-status.ts
sebastien-comeau: Because you are not suposed to return the response but mutatate it with the handler function. So I mutate the reponse then early return.

pages/api/check-status.ts
sebastien-comeau: I don't think it's "janky" and error prone. @gregory-j-baker what's your point of view?

pages/api/check-status.ts
sebastien-comeau: Again, because you are not suposed to return the response but mutatate it with the handler function. So I mutate the reponse then early return.

__tests__/api/check-status.test.ts
sebastien-comeau: Good point, will remove.

__tests__/api/check-status.test.ts
sebastien-comeau: Good point, forgot to remove the export.

pages/api/check-status.ts
sebastien-comeau: Will do, thanks.

sebastien-comeau: I think this api handler could be revisit to remove some throwing of ApiError. Will take a note.

pages/api/check-status.ts
gregory-j-baker: My first thought when I saw this was that there is definitely some jank here.. the framework should provide this functionality, and having to do it by hand by wrapping `req.url` in `new URL()` just looks weird.

And resolving the query params using strings is definitely prone to typos; but even if this were somehow strongly typed, those types would end up just being hand-typed types (man that's a lot of _type_). My point is that resolving query params is prone to error no matter how you do it; someone is going to have to type these strings out at some point, whether it be in a `searchParams.get()` or a `interface CheckStatusRequestType { .. }`.

I don't have a strong point of view either way.. this code is ugly (not your fault), but it works. It would be nice if there was a cleaner way to do it.

lib/CheckStatusHook.ts
s-laugh: It's kind of nice, and easy to re-write the `URLSearchParams` ourselves if it requires a whole package to be installed just for that.

pages/api/check-status.ts
s-laugh: ahh, my eyes have lied to me - I was sure I saw returns everywhere in the examples, but they aren't there at all...

pages/api/check-status.ts
s-laugh: That is exactly my feeling of it too - but I don't think it's worth blocking the PR over it. Maybe this is something we could build a little custom function for later to handle in a slightly "cleaner", more predictable fashion - even through in some logging ;).

PR #84, opened 2022-10-19
Changed default style to match canada.ca default button background co…

Maxim-Lam:
…lor and applied it to modal cancel button.

### Description

List of proposed changes:

Changed the background color of the default styling to match canada.ca one.

Applied the style to the modal's "no" button

Also fix the focus issue bug

Visit status page, look at cancel button and put it in focus by either tabbing to it or hold left click on it. Text should be visible now.

Select cancel and do the same thing with the no button

### DIFF
--- a/components/ActionButton.tsx
+++ b/components/ActionButton.tsx
@@ -20,11 +20,11 @@ const ActionButton: FC<ActionButtonProps> = ({
   switch (style) {
     case 'primary':
       classStyle +=
-        'border-blue-dark bg-blue-dark text-basic-white focus:bg-blue-normal hover:bg-blue-normal active:bg-blue-active'
+        'border-blue-dark bg-blue-dark text-basic-white focus:bg-blue-normal hover:bg-blue-normal active:bg-blue-active shadow-sm'
       break
     default:
       classStyle +=
-        'border-gray-default bg-gray-light text-blue-light hover:border-gray-deep hover:bg-gray-dark'
+        'border-gray-dark bg-gray-normal text-blue-light hover:border-gray-deep hover:bg-gray-dark focus:text-blue-light focus:bg-gray-dark shadow-sm'
       break
   }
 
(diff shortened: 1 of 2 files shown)


components/ActionButton.tsx
```diff
@@ -20,11 +20,11 @@ const ActionButton: FC<ActionButtonProps> = ({
   switch (style) {
     case 'primary':
       classStyle +=
-        'border-blue-dark bg-blue-dark text-basic-white focus:bg-blue-normal hover:bg-blue-normal active:bg-blue-active'
+        'border-blue-dark bg-blue-dark text-basic-white focus:bg-blue-normal hover:bg-blue-normal active:bg-blue-active shadow-sm'
       break
     default:
       classStyle +=
-        'border-gray-default bg-gray-light text-blue-light hover:border-gray-deep hover:bg-gray-dark'
+        'border-gray-dark bg-gray-normal text-blue-light hover:border-gray-deep hover:bg-gray-dark focus:text-blue-light focus:bg-gray-dark shadow-sm'
```
krischarbonneau, line 27: Going off the design pattern for the secondary button [here](https://design.canada.ca/common-design-patterns/buttons.html) we can see that the button has an outset border. There isn't an outset border style in tailwind but we can simulate that by setting borders on individual sides like the below suggestion.

```suggestion
        'border-gray-dark bg-gray-normal text-blue-light hover:bg-gray-dark hover:border-l-gray-deep hover:border-t-gray-deep focus:text-blue-light focus:bg-gray-dark shadow-sm border-r-gray-500 border-b-gray-500'
```

Maxim-Lam: very nice. I was bummed that there wasn't a way to do it like 'border-r-b...'. good suggestion

krischarbonneau: Tailwind is a great tool that makes styling easy, but it still can't hold a candle to what you can do with conventional CSS

krischarbonneau (approved):
Looks good!

components/ActionButton.tsx
Maxim-Lam: omg i agree

PR #85, opened 2022-10-19
add response, confirmation

s-laugh:
### Description

List of proposed changes:

- Add Response for email sent 
- Moved contact info to common location
- Removed custom hook _As I was debugging problems getting it to work, as I got this working but couldn't go back to the custom hook without problems. It also felt cleaner & easier to read, so I left it._

If the fields on the email page are valid, submit should give you the response.
If the API fails for some unknown reason (can be forced by change code to return a non-200 response) then the error page should load.

### Additional Notes

I'll be opening a bug for the Error pages - the links back to home break (looks to be something with i18next)


gregory-j-baker:
Re: the custom hook

This blog post makes a good point on why custom hooks are the preferred way to handle data fetching: https://tkdodo.eu/blog/practical-react-query#create-custom-hooks (I linked to this yesterday in our dev chat, btw)

I am suggesting you at least consider adding it back because I feel that it's a beneficial pattern that will make the application more maintainable.

sebastien-comeau:
I fiddle a little bit with the mutation hook and here's what it would looks like:

```ts
// useEmailEsrf.ts
import { useMutation, UseMutationOptions } from '@tanstack/react-query'
import { ApiError } from 'next/dist/server/api-utils'
import { EmailEsrfRequestBody } from '../pages/api/email-esrf'

const useEmailEsrf = (
  options?: UseMutationOptions<void, ApiError, EmailEsrfRequestBody>
) => {
  return useMutation<void, ApiError, EmailEsrfRequestBody>(async (body) => {
    const response = await fetch('/api/email-esrf', {
      method: 'POST',
      body: JSON.stringify(body),
    })

    if (!response.ok) {
      throw new ApiError(response.status, response.statusText)
    }
  }, options)
}

export default useEmailEsrf
```

and when calling it in `email.tsx` page:

```tsx
const {
    isLoading: isEmailEsrfLoading,
    isSuccess: isEmailEsrfSuccess,
    error: emailEsrfError,
    mutate: emailEsrf,
  } = useEmailEsrf()
```

and you throw the `ApiError` exception if it's not undefined like we do in `status.tsx` page:

```tsx
if (emailEsrfError) throw emailEsrfError
```

s-laugh:
> and you throw the `ApiError` exception if it's not undefined like we do in `status.tsx` page:

So I  tried throwing it, but I never actually got an error, the result still rendered like nothing happened (the throw was caught and handled by the browser)

sebastien-comeau:
> > and you throw the `ApiError` exception if it's not undefined like we do in `status.tsx` page:
> 
> So I tried throwing it, but I never actually got an error, the result still rendered like nothing happened (the throw was caught and handled by the browser)

Yeah, I tweaked the api endpoint and now it works. Here's the the endpoint code:

```ts
import { NextApiRequest, NextApiResponse } from 'next'
import logger from '../../lib/logger'

export interface EmailEsrfRequestBody {
  dateOfBirth: string
  email: string
  firstName: string
  lastName: string
}

export default async function handler(
  req: NextApiRequest,
  res: NextApiResponse<string>
) {
  if (req.method !== 'POST') {
    res.status(405).send(`Invalid request method ${req.method}`)
    return
  }

  try {
    const response = await (process.env.PASSPORT_STATUS_API_BASE_URI
      ? emailEsrfApi(req.body)
      : emailEsrfMock())
    res.status(response.status).send(response.statusText)
  } catch (error) {
    logger.error(error)
    res.status(500).send('Something went wrong.')
  }
}

const emailEsrfApi = async (rqBody: any) => {
  return fetch(
    `${process.env.PASSPORT_STATUS_API_BASE_URI}/api/v1/electronic-service-requests/`,
    { method: 'POST', body: rqBody }
  )
}

const emailEsrfMock = async () => {
  //unless there is an error in the network communication or bad data format,
  //we expect the response to always be a 202 to protect the users data
  return new Response(null, { status: 500, statusText: 'Email sent if found' })
}
```

sebastien-comeau:
> @sebastien-comeau No matter what I try when using your suggestions, the form will not submit.

Here's a branch I made from this PR branch. My original suggestions was working with mock but I didn't try it with the API. I fiddle a little bit this morning and both mock and API works as attented.

https://github.com/DTS-STN/passport-status/tree/confirmationEmailWithCustomMutationHook

s-laugh:
The happy path looks to work with the added hook. But I can't approve my own PR :)

sebastien-comeau (approved):

PR #86, opened 2022-10-19
Created contact us page

krischarbonneau:
### Description

This PR adds a `Contact Us` page with links to contact the passport program and other resources as per the task created by the BA in our backlog. Currently there is no way to navigate to this page aside from manually doing so (this will come once we get more clarity on task 1054 and may result in some text on this page changing).

1. Pull in branch
2. Type `npm run dev`
3. Navigate to `/contact`
4. Ensure content and links show up properly in both EN and FR & open resource in a new tab


components/LinkButton.tsx
```diff
@@ -6,12 +6,21 @@ export interface LinkButtonProps {
...
+  text,
+  id,
+  lang,
+  external,
+}) => {
   return (
     <Link href={href} passHref>
       <a
+        target={external ? '_blank' : ''}
+        rel={external ? 'noopener noreferrer' : ''}
```
s-laugh, line 23: Could use && here

pages/contact.tsx
```diff
@@ -0,0 +1,55 @@
...
+            external
+          ></LinkButton>
+        </div>
+        <div id="bookAppointment">
+          <LinkButton
+            href={t('links.bookAppointment.link')}
+            text={t('links.bookAppointment.text')}
+            external
+          ></LinkButton>
+        </div>
```
s-laugh, line 40: I'm pretty sure these are the same links found on our other pages like the status results (although maybe one if the links isn't right) Wondering if it makes sense to re-use the component (you do have it styled differently, so a class(es) could be passed into the component)
https://github.com/DTS-STN/passport-status/blob/main/pages/status.tsx#L131
Do note, that I'm suggesting a change to how the text for it is extracted from common https://github.com/DTS-STN/passport-status/pull/85/files#diff-9679a3f1633cdb4d4062246a217d7eb49bdcc6b5889b090e82e1ddc06dc76518R79

s-laugh (commented):
Looks good! I'm not sure about the button styling for the links...

pages/contact.tsx
krischarbonneau: No issue switching over to using the LinkSummary component, I also agree with your second point/change.

components/LinkButton.tsx
krischarbonneau: Will be switching to using the LinkSummary component but I'll still update this for external links if for whatever reason we want to link to an external site via a button in the future.

cypress/e2e/contact.cy.js
```diff
@@ -8,16 +8,24 @@ describe('contact page loads', () => {
...
+      cy.get('[data-cy=toggle-language-link]').should('contain.text', 'English');
+  
     })
 
-    it('should display the button that links to appointment booking resource',()=>{
-        cy.get(`#bookAppointment`).should('be.visible')
+    it('should display a list of links',()=>{
+        cy.get(`#contactLinks`).contains("Contact the Passport Program")
+        cy.get(`#contactLinks`).contains("Find a passport service location near you")
+        cy.get(`#contactLinks`).contains("Appointment booking")
```
s-laugh, line 28: Not a huge fan of testing text as it can change breaking the test, I think testing that there is at least one link should achieve what we want from this test.

s-laugh (approved):
Great work!

PR #87, opened 2022-10-20
1122 error page links

Maxim-Lam:
### Description

List of proposed changes:

error page "home page" link now returns user to the home page.

enter incorrect url. localhost:xxxx/xyz. Click home page on both sections of the text to make sure it return to home.

### DIFF
--- a/pages/404.tsx
+++ b/pages/404.tsx
@@ -38,7 +38,7 @@ const Custom404 = () => {
         <ul>
           <li>
             Retournez à la{' '}
-            <Link href="/">
+            <Link href="/" locale="default">
               <a className="text-cyan-600 underline">page {"d'accueil;"}</a>
             </Link>
           </li>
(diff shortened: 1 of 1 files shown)


Maxim-Lam:
> Have you an explanation on why this is happening. I tried to figure out locally after seeing your PR but can't found the reason. I thought maybe the middleware was the issue but it is not the case. It works as attended.

Hey @sebastien-comeau I know this is late so you won't see this till the morning but it seems as though the error page is by default set to use 'en' locale and therefore the `<Link href="/">` returns localhost:3000/en which does not exist. Setting the local to default for the links on the error page returns the user to the home page here.

sebastien-comeau:
@Maxim-Lam, your solution works but a link to `localhost:3000/en` should work because if I type `localhost:3000/en` in the browser the middleware redirect me to `localhost:3000/` so why a link in 404 with `localhost:3000/en` doesn't? I will approve your PR but will try replicate this in a vanilla Next.js app. Thanks :)

sebastien-comeau (approved):

PR #89, opened 2022-10-24
Update types to match passport-status-api NIEM changes

sebastien-comeau:
### Description

List of proposed changes:

- Update types to match passport-status-api NIEM changes

Set env. variable `PASSPORT_STATUS_API_BASE_URI=https://passport-status-api.dev.dev-rhp.dts-stn.com` in `.env.local` file then try the following inputs:

esrf: 002D85ED
givenName: Yanis
surname: Piérre
dob: 1972-07-29

Status results should be: "1" translated to "Approved"


krischarbonneau (approved):
Works well.

PR #99, opened 2022-10-27
Added privacy/emailterm pages into the workflow. Missing actual data.

Maxim-Lam:
### Description

##Opinion needed, not to be merged YET.

I added the privacy and email terms page before the consent page. I took a little creative liberty with the flow and would like your opinion.

I went with a sort of multi-step form in the sense of "next -> next -> I agree -> [ redirect to Form Page]"; Should I add the consent portion into this flow or have this flow match the consent page.

I know it may sound complicated but you will understand once you play with it.

List of proposed changes:

- Added privacy and email terms page
- modify flow

in the landing page select "I don't have my ESRF number"

Follow the flow

### DIFF
--- a/pages/privacy.tsx
+++ b/pages/privacy.tsx
@@ -20,7 +20,7 @@ const Privacy: FC = () => {
       <h1 className="mb-4">{t('header')}</h1>
       <div className="ml-auto mr-auto p-3">
         {!agree && (
-          <span id="terms">
+          <div id="terms">
             <h2 className="my-14">{t('terms-text')}</h2>
             <div
               id="termsBtnGrp"
@@ -37,11 +37,11 @@ const Privacy: FC = () => {
                 onClick={() => Router.push('/contact')}
               />
             </div>
-          </span>
+          </div>
         )}
 
         {agree && (
-          <span id="privacy">
+          <div id="privacy">
             <h2 className="my-14">{t('privacy-text')}</h2>
             <div
               id="privacyBtnGrp"
@@ -58,7 +58,7 @@ const Privacy: FC = () => {
                 style="primary"
               />
             </div>
-          </span>
+          </div>
         )}
       </div>
     </Layout>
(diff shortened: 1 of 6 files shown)


pages/privacy.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  return (
+    <Layout
+      meta={t('common:meta', { returnObjects: true })}
+      header={t('common:header', { returnObjects: true })}
+      footer={t('common:footer', { returnObjects: true })}
+    >
+      <h1 className="mb-4">{t('header')}</h1>
+      <div className="max-w-4xl overflow-hidden ml-auto mr-auto p-3">
+        <span id="terms" style={{ display: !agree ? 'block' : 'none' }}>
```
sebastien-comeau, line 22: Use a `div` instead of `span` and conditionally render it based on `agree` state.

```tsx
{!agree && <div id="terms">...</div>}
```

pages/privacy.tsx
```diff
@@ -0,0 +1,67 @@
...
+              style="primary"
+              onClick={() => setAgree(true)}
+            />
+            <ActionButton
+              text={t('disagree')}
+              onClick={() => Router.push('/contact')}
+            />
+          </div>
+        </span>
+        <span id="privacy" style={{ display: agree ? 'block' : 'none' }}>
```
sebastien-comeau, line 39: Use a `div` instead of `span` and conditionally render it based on `agree` state.

```tsx
{agree && <div id="privacy">...</div>}
```

sebastien-comeau (commented):
Trivial changes but we will need to have an open discussion on how this privacy and consent can be forced to the user before accessing the form. The problem we have at the moment is nothing can block the user to go directly to the form. The user can bypass each steps.

Maxim-Lam:
@sebastien-comeau  I will make the changes you suggested as well as kris' recommendation to make the style match the consent page.

sebastien-comeau (approved):
Looks go to me. Still need to have an open discussion about those steps that can be easily skipped.

PR #100, opened 2022-10-27
Update missing translations

krischarbonneau:
### Description
Was pointed out that our Date modified component wasn't translating. Turns out the reason was because the key was different for French. I've changed it to be in line with English and it now properly translates. I also noticed that the keys were different for the buttons on the consent page so I've added that fix in this PR as well. I also wrapped the `<dl>` in the DateModified component with `<time>` as per Greg's suggestion.

1. Pull in branch
2. Type `npm run dev`
3. Navigate to `/consent`
4. Switch languages and ensure both the date modified text and the text for the 2 buttons are translated.

### DIFF
--- a/components/DateModified.tsx
+++ b/components/DateModified.tsx
@@ -13,16 +13,18 @@ export interface DateModifiedProps {
  */
 const DateModified: FC<DateModifiedProps> = ({ id, text }) => {
   return (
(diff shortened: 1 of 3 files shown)


Maxim-Lam (approved):
Like a charm!

PR #106, opened 2022-10-31
Prevent prevents users from accessing form till the privacy terms are…

Maxim-Lam:
### Description

List of proposed changes:

prevents users from completing the form before accepting email and privacy terms

from landing page change the url from localhost:XXXX/en/email before accepting and consenting to privacy and email terms

navigate to terms

accept

voila

### DIFF
--- a/middleware.ts
+++ b/middleware.ts
@@ -1,4 +1,5 @@
 import { NextRequest, NextResponse } from 'next/server'
+import { hasCookie } from 'cookies-next'
 
 //regex to check if there's an extension in the path, ie .jpg
 const PUBLIC_FILE = /\.(.*)$/
@@ -22,5 +23,12 @@ export async function middleware(req: NextRequest) {
     req.nextUrl.pathname === '/'
   ) {
     return NextResponse.redirect(new URL(`/`, req.url))
+  }
+
+  if (
+    req.cookies.get('agreed-to-terms') != 'undefined' &&
+    req.nextUrl.pathname === '/email'
+  ) {
+    return NextResponse.redirect(new URL('/privacy', req.url))
   }
 }
--- a/middleware.ts
+++ b/middleware.ts
@@ -26,9 +26,11 @@ export async function middleware(req: NextRequest) {
   }
 
   if (
-    req.cookies.get('agreed-to-terms') != 'undefined' &&
+    req.cookies.get('agreed-to-email-esrf-terms') !== 'true' &&
     req.nextUrl.pathname === '/email'
   ) {
-    return NextResponse.redirect(new URL('/privacy', req.url))
+    return NextResponse.redirect(
+      new URL(`/${req.nextUrl.locale}/privacy`, req.url)
+    )
   }
 }
--- a/middleware.ts
+++ b/middleware.ts
@@ -26,8 +26,8 @@ export async function middleware(req: NextRequest) {
   }
 
   if (
-    req.cookies.get('agreed-to-email-esrf-terms') !== 'true' &&
-    req.nextUrl.pathname === '/email'
+    req.nextUrl.pathname === '/email' &&
+    req.cookies.get('agreed-to-email-esrf-terms') !== 'true'
   ) {
     return NextResponse.redirect(
       new URL(`/${req.nextUrl.locale}/privacy`, req.url)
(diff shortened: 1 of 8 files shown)


sebastien-comeau:
I suggest to write `useSessionStorage` hook that will handle reading and saving in session storage then you can write your own custom hook that return `useSessionStorage`. The custom hook will set the session storage key and the default value. It should also ease unit tests when mocking the custom hook return value.

```ts
export useAgreedToTermsStorage = () => (useSessionStorage("agreed-to-terms", false));
``

Here's an example of `useSessionStorage` hook.
https://usehooks-ts.com/react-hook/use-session-storage

We can either use it from this library or write it. This hook support SSR which is critical with Next.js. When you build the application every static pages will be rendered server side. It's why you need to use `useEffect` to read the value from session storage of the client.

File: pages/email.tsx [hunk 1]
```diff
@@ -60,113 +62,145 @@ export default function Email() {
...
+                        ]}
+                      />
+                    </div>
+                  </div>
+                </form>
+              </>
+            )
+          }
+        } else {
+          return (
```
sebastien-comeau, line 193: This return will be call as default rendering because the session storage can only be read client-side. So we might see the screen flicker because it will load with `false` by default then `useEffect` will try to read the value then render accordingly.

sebastien-comeau: If ever a search engine crawl the website this page will be indexed with this return component. It makes me rethink how do we want to handle the user consents thing. Maybe using a cookie is the way to go because we can redirect server side if the user didn't consent based on cookie.

sebastien-comeau: Tested it and if someone already agreed and refresh the email page then you see the screen flickered from `page-unauth` to `email-form`. I might be a little bit finiky.

pages/email.tsx
```diff
@@ -66,107 +67,121 @@ export default function Email() {
...
-                    type: 'button',
-                  },
-                  {
-                    text: t('common:cancel-modal.no-button'),
-                    onClick: () => setModalOpen(!modalOpen),
-                    style: 'default',
-                    type: 'button',
-                  },
-                ]}
+      {(() => {
```
sebastien-comeau, line 70: Change this to inline-if conditional rendering.

middleware.ts
```diff
@@ -23,4 +24,13 @@ export async function middleware(req: NextRequest) {
   ) {
     return NextResponse.redirect(new URL(`/`, req.url))
   }
+
+  if (
```
sebastien-comeau, line 28: Change the order of the AND condition to be more readable
```js
// redirect for email page if user didn't agreed on email esrf terms
if (req.nextUrl.pathname === '/email' && req.cookies.get('agreed-to-email-esrf-terms') !== 'true') { ... }
```

pages/consent.tsx
```diff
@@ -18,7 +21,16 @@ const Consent: FC = () => {
       <h2 className="my-14">{t('description')}</h2>
       <div className="flex justify-center flex-wrap text-xl gap-4">
         <div id="yes-button">
-          <LinkButton text={t('yes-button')} href="/email" />
+          <ActionButton
+            text={t('yes-button')}
+            onClick={() => {
```
sebastien-comeau, line 26: Move this handle in his own function.

```js
const handleOnAgreeClick: MouseEventHandler<HTMLButtonElement> = useCallback((e) => {
      e.preventDefault()
      setCookie('agreed-to-email-esrf-terms', 'true', { sameSite: true })
      router.push('/email')
    }, [])
```

pages/email.tsx
```diff
@@ -6,7 +6,7 @@ import { useTranslation } from 'next-i18next'
 import { serverSideTranslations } from 'next-i18next/serverSideTranslations'
 import Layout from '../components/Layout'
 import { EmailEsrfRequestBody } from './api/email-esrf'
-import { useMemo, useState } from 'react'
+import { useEffect, useMemo, useState } from 'react'
```
sebastien-comeau, line 9: Remove unused `useEffect`

pages/email.tsx
```diff
@@ -16,6 +16,7 @@ import ActionButton from '../components/ActionButton'
 import Modal from '../components/Modal'
 import LinkSummary, { LinkSummaryItem } from '../components/LinkSummary'
 import useEmailEsrf from '../lib/useEmailEsrf'
+import LinkButton from '../components/LinkButton'
```
sebastien-comeau, line 19: Remove unused `LinkButton`

sebastien-comeau (commented):
Trival changes

PR #108, opened 2022-11-02
1220 i18next nesting

sebastien-comeau:
### Description

List of proposed changes:

- Remove i18next nesting because it causes problem when reading and trying to figure out the end results of the string. It also cause problem with reusing key that have string case for a specific situtation but needs to be in a different case in another place.
- Revisited all the locales JSON files to have consistenty with terms and case.

Load the application and ensure each page display as it should in both languages.

### DIFF
--- a/public/locales/en/status.json
+++ b/public/locales/en/status.json
@@ -12,10 +12,10 @@
   "description": "Use this service with your file number (ESRF) to check the status of your passport application.",
   "esrf": {
     "error": {
-      "length": "The file number (ESRF) must be 8 characters long.",
-      "required": "The file number (ESRF) is required."
+      "length": "The file number must be 8 characters long.",
+      "required": "The file number is required."
     },
-    "label": "File number (ESRF)"
+    "label": "File number"
   },
   "given-name": {
     "error": {
--- a/public/locales/en/status.json
+++ b/public/locales/en/status.json
@@ -2,7 +2,7 @@
   "date-of-birth": {
     "error": {
       "current": "The date of birth must be in the past.",
-      "invalid": "The date of birth must be in a valid format of «\u00a0yyyy-MM-dd\u00a0».",
+      "invalid": "The date of birth must be valid in the following format: YYYY-MM-DD (Year-Month-Day).",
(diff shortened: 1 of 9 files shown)


public/locales/en/status.json
```diff
@@ -1,27 +1,27 @@
 {
   "date-of-birth": {
     "error": {
-      "current": "The $t(date-of-birth.label) $t(common:error-messages.in-past)",
-      "invalid": "The $t(date-of-birth.label) $t(common:error-messages.valid-format, { \"format\": \"dd/mm/yyyy\" })",
-      "required": "The $t(date-of-birth.label) $t(common:error-messages.required)"
+      "current": "The date of birth must be in the past.",
+      "invalid": "The date of birth must be in a valid format of «\u00a0yyyy-MM-dd\u00a0».",
```
krischarbonneau, line 5: Aren't «» only used as quotations in French? Should this not be use `""` here?

sebastien-comeau: You might be right, thanks.

Will adjust and will see what is done on Canada.ca websites.

sebastien-comeau: Adjusted based on an IRCC guide:

- [English](https://www.canada.ca/en/immigration-refugees-citizenship/services/application/application-forms-guides/guide-5218-request-amend-record-landing-confirmation-permanent-residence-valid-temporary-resident-documents.html#:~:text=date%20of%20birth-,in%20the%20following%20format%3A%20YYYY%2DMM%2DDD%20(Year%2DMonth%2DDay),-Question%208)
- [French](https://www.canada.ca/fr/immigration-refugies-citoyennete/services/demande/formulaires-demande-guides/guide-5218-demande-modification-fiche-relative-droit-etablissement-confirmation-residence-permanente-documents-resident-temporaire-valides.html#:~:text=date%20de%20naissance-,selon%20le%20format%20suivant%C2%A0%3A%20AAAA%2DMM%2DJJ%20(Ann%C3%A9e%2DMois%2DJour).,-Question%208)

krischarbonneau (approved):
Looks good!

PR #110, opened 2022-11-04
Add Reusable ErrorPage component, update error pages

krischarbonneau:
### Description
This PR adds a reusable `ErrorLayout` component with header & footer to match Canada.ca. I've added meta data using our existing component but that does not include OG or DCMI as those should really be added to the `MetaData` component in another PR.

1. Pull in branch
2. Type `npm run dev`
3. Navigate to `/404` and ensure the page renders properly. Test at mobile screen width to confirm that the top of page anchor appears and works properly

### Additional Notes
Due to the current official workaround for adding the default locale to the path being done through middleware, the user is automatically redirected to `/en/404`. We don't necessarily want this since the page is meant to be bilingual, similar to the splash page, but the workaround that was used for the splash page doesn't work for the error pages and I'm not entirely sure why. Does next handle these error pages differently? Worth noting that while you can specify at the top of the middleware that if the pathname starts with `/404` to not redirect to `/en/404`, this does work but if you manually navigate to `/foo` for example, you still get redirected to `/en/404`.


components/ErrorPage.tsx
```diff
@@ -0,0 +1,118 @@
+import { FC } from 'react'
+import Link from 'next/link'
+import MetaData from './MetaData'
+
+export interface ErrorPageProps {
+  statusCode?: number | undefined
+}
+
+const ErrorPage: FC<ErrorPageProps> = ({ statusCode }) => {
```
sebastien-comeau, line 9: This component should be `ErrorLayout.tsx`. The error layout component would be used in `404.tsx` and `_error.tsx`. It would have `children: React.ReactNode` as property. It will be reponsible for the error pages layout. The `404.tsx` and `_error.tsx` page would be the same as before and bilingual. `MetaData` will be used in `404.tsx` and `_error.tsx`.

- `404.tsx` [english](https://www.canada.ca/en/errors/404.html) and [french](https://www.canada.ca/fr/erreurs/404.html) from Canada.ca
- `_error.tsx` for 503 [english](https://www.canada.ca/en/errors/503.html) and [french](https://www.canada.ca/fr/erreurs/503.html) from Canada.ca
- `_error.tsx` for 500 and other status codes [english](https://www.canada.ca/en/errors/500.html) and [french](https://www.canada.ca/fr/erreurs/500.html) from Canada.ca
- `_error.tsx` for client error (when status code is undefined) would have is own content. I don't know what to use content though.

sebastien-comeau (commented):
Code refactoring

components/ErrorPage.tsx
krischarbonneau: Ah okay, I misunderstood the task then when it was brought up during standup. I'll make these changes.

sebastien-comeau (changes requested):
I've played wit your change and I though it would be easier to create a PR #113 to this branch so you can take a look of my recommendations.

sebastien-comeau (approved):

PR #117, opened 2022-11-09
Updated ESRF validation

krischarbonneau:
### Description
This PR adds additional validation to the ESRF form field based on newest requirements. In these requirements it is stated that the ESRF must begin with a letter (Examples of possible formats: A1234567 or AA123456 or A123456 or AA12345). The requirements state that we only need to check for the first character, so we don't have to check for all possible formats, just that it begins with a letter. I've altered the mocks to reflect this for the time being until new mocks are generated just so Cypress tests can pass. 

1. Pull in branch
2. Type `npm run dev`
3. Navigate to `/status`
4. Enter `12345678` and confirm that an error pops up stating the proper ESRF format


public/locales/fr/status.json
```diff
@@ -13,7 +13,8 @@
   "esrf": {
     "error": {
       "length": "Le numéro de dossier doit comporter 8 caractères.",
-      "required": "Le numéro de dossier est obligatoire."
+      "required": "Le numéro de dossier est obligatoire.",
+      "starts-with-letter": "Le numéro de dossier doit commencer par une lettre (p. ex. A1234567)"
```
sebastien-comeau, line 17: A period is missing at the end.

sebastien-comeau (commented):
Minor tweak to be done.

sebastien-comeau (approved):

PR #118, opened 2022-11-10
Add manifest number to check status api response

sebastien-comeau:
### Description

List of proposed changes:

- Add manifest number in the api check status reponse

### What to test for/How to test

Check the status of an applicantion that have a `"Manifest Number"` identification.

```json
{
  "_embedded": {
    "GetCertificateApplicationResponse": [
      {
        "id": "e06b819b-3f83-327b-b6ad-140873ffffb5",
        "CertificateApplication": {
          "CertificateApplicationIdentification": [
            {
              "IdentificationCategoryText": "Application Register SID",
              "IdentificationID": "5fb36a76-0527-3f40-8381-db7fb821f76d"
            },
            {
              "IdentificationCategoryText": "File Number",
              "IdentificationID": "00010D5D"
            },
            {
              "IdentificationCategoryText": "Manifest Number",
              "IdentificationID": "ECF9A379"
            }
          ],
          "CertificateApplicationStatus": {
            "StatusCode": "3"
          }
        }
      }
    ]
  }
}
```

### Additional Notes

The manifest number coming from the API represent either Canada Post or Fedex tracking number.

image

PR #120, opened 2022-11-10
Update status codes

krischarbonneau:
### Description
This PR updates the status codes and corresponding messages & descriptions based on the table in this PR : https://github.com/DTS-STN/passport-status-api/pull/46

This also makes use the manifest number, which was added to the check status response in #118, to display the user's tracking number if their status is either a 3 or a 4 (issued and shipped via Canada Post or Fedex). There's some additional logic to determine whether the status code is either 3 or 4 since there are some records (at least in the mocks) that have a manifest number even though the user's passport hasn't been shipped yet (e.g. we have a record with the status code 5, which is rejected, that has a manifest number)

1. Pull in branch
2. Type `npm run dev`
3. Navigate to `/status`
4. Enter the following mock data:
```
ESRF: A0890196
Given Name: Manon
Surname: Le roux
DoB: 1987-02-24
```
5. Press submit and confirm that the tracking number is shown (and it says it's a Canada Post tracking number)
6. Try with other mock data to test out the other status codes (dm if you don't know where to find this)

### Additional Notes
In `status.tsx` CheckStatusReponse should be CheckStatusRe**s**ponse


pages/status.tsx
```diff
@@ -112,15 +112,26 @@ const Status: FC = () => {
...
+                      {t(`status.${checkStatusReponse.status}.label`, {
+                        defaultValue: checkStatusReponse.status,
+                      })}
+                    </strong>
+                    .
+                  </p>
+                  <p>{t(`status.${checkStatusReponse.status}.description`)}</p>
+                  {/* We only want to show the tracking # if the status is issued and shipped via Canada Post or Fedex. Currently, some records have a manifest # but are not one of these codes (ie. 99) */}
+                  {checkStatusReponse.manifestNumber &&
+                    ['3', '4'].indexOf(checkStatusReponse.status) >= 0 && (
```
stefan-oconnell, line 128: Sorry my javascript is rusty, but could this be: ['3', '4'].includes(checkStatusResponse.status)? Or is that not a thing on vanilla arrays?

sebastien-comeau: I agree with Stefan with using `[].includes(value)`.

Use TS enum instead of harcoding `"3"` and `"4"`. The enum can be added to `/lib/type.ts`.
https://www.typescriptlang.org/docs/handbook/enums.html#string-enums

Here's the enum of status code that match the API:

```ts
export enum StatusCode {
    APPLICATION_NO_LONGER_MEETS_CRITERIA = "99".
    FILE_BEING_PROCESSED = "1",
    NOT_ACCEPTABLE_FOR_PROCESSING = "5",
    PASSPORT_ISSUED_READY_FOR_PICKUP = "2",
    PASSPORT_ISSUED_SHIPPING_CANADA_POST = "3",
    PASSPORT_ISSUED_SHIPPING_FEDEX = "4"
}
```

Remove the comment when It's done.

pages/status.tsx
```diff
@@ -112,15 +112,26 @@ const Status: FC = () => {
...
-                      defaultValue: checkStatusReponse.status,
-                    })}
-                  </strong>
-                  .
-                </p>
+                <div>
+                  <p className="mb-6 text-2xl">
+                    {`${t('status-is')} `}
+                    <strong id="response-status">
+                      {t(`status.${checkStatusReponse.status}.label`, {
```
sebastien-comeau, line 119: Remove `defaultValue` option. I prefer to display `status.99,label` if the key is missing. I know that I am the one who added  `defaultValue` at the beginning. 😄

sebastien-comeau (commented):
Minor tweaks. English and French conent need to be reviewed and it can be done later.

lib/types.ts
```diff
@@ -32,5 +32,14 @@ export interface PassportStatusesCertificateApplicationIdentification {
 }
 
 export interface PassportStatusesCertificateApplicationStatus {
-  StatusCode: string
+  StatusCode: StatusCode | string
```
sebastien-comeau, line 35: Set back `StatusCode` type to `string`.

pages/status.tsx
```diff
@@ -112,15 +113,26 @@ const Status: FC = () => {
...
+                  <p className="mb-6 text-2xl">
+                    {`${t('status-is')} `}
+                    <strong id="response-status">
+                      {t(`status.${checkStatusReponse.status}.label`)}
+                    </strong>
+                    .
+                  </p>
+                  <p>{t(`status.${checkStatusReponse.status}.description`)}</p>
+                  {checkStatusReponse.manifestNumber &&
+                    [
```
sebastien-comeau, line 126: Use the enum value of type string here.

```tsx
[`${StatusCode.PASSPORT_ISSUED_SHIPPING_CANADA_POST}`,
`${StatusCode.PASSPORT_ISSUED_SHIPPING_FEDEX}`]
```

sebastien-comeau (approved):
Looks good to me.

> Locales needs to be reviewed later in both languages.

PR #125, opened 2022-11-14
Update esrf validation and disallow spaces

krischarbonneau:
### Description
This PR updates the validation schema for the ESRF field to be a max of 8 characters, instead of a required length of 8 characters. I've also updated the `onChange` for the ESRF input to not allow spaces.

1. Pull in branch
2. Type `npm run dev`
3. Navigate to `/status`
4. Confirm that you are not able to add spaces to the field and that any entry starting with a letter and having a length that does not exceed 8 characters is considered valid.

### DIFF
--- a/pages/status.tsx
+++ b/pages/status.tsx
@@ -42,7 +42,7 @@ const Status: FC = () => {
         .max(new Date(), 'date-of-birth.error.current'),
       esrf: Yup.string()
         .required('esrf.error.required')
-        .length(8, 'esrf.error.length')
+        .max(8, 'esrf.error.length')
         .trim()
         .matches(/^[A-Za-z]/, 'esrf.error.starts-with-letter'),
       givenName: Yup.string().required('given-name.error.required'),
@@ -199,7 +199,12 @@ const Status: FC = () => {
               id="esrf"
               name="esrf"
               label={t('esrf.label')}
-              onChange={formik.handleChange}
+              onChange={(event) =>
+                formik.setFieldValue(
+                  event.target.name,
+                  event.target.value.replace(/[^a-z0-9]/gi, '')
+                )
+              }
               value={formik.values.esrf}
               errorMessage={formik.errors.esrf && t(formik.errors.esrf)}
               textRequired={t('common:required')}
--- a/public/locales/en/status.json
+++ b/public/locales/en/status.json
@@ -12,7 +12,7 @@
   "description": "Use this service with your file number (ESRF) to check the status of your passport application.",
   "esrf": {
     "error": {
-      "length": "The file number must be 8 characters long.",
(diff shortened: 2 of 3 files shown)


public/locales/en/status.json
```diff
@@ -12,7 +12,7 @@
   "description": "Use this service with your file number (ESRF) to check the status of your passport application.",
   "esrf": {
     "error": {
-      "length": "The file number must be 8 characters long.",
+      "length": "The file number can be a maximum of 8 characters long.",
```
sebastien-comeau, line 15: I think this better:

`The file number can have a maximum of 8 alphanumeric characters.`

public/locales/fr/status.json
```diff
@@ -12,7 +12,7 @@
   "description": "Utilisez ce service avec votre numéro de dossier (FEDS) pour vérifier le statut de votre demande de passeport.",
   "esrf": {
     "error": {
-      "length": "Le numéro de dossier doit comporter 8 caractères.",
+      "length": "Le numéro de dossier peut comporter un maximum de 8 caractères.",
```
sebastien-comeau, line 15: I think this is better:

`Le numéro de dossier peut avoir un maximum de 8 caractères alphanumériques.`

sebastien-comeau (commented):
Minor tweak and feel free to debate. 😃

pages/status.tsx
```diff
@@ -199,7 +199,12 @@ const Status: FC = () => {
               id="esrf"
               name="esrf"
               label={t('esrf.label')}
-              onChange={formik.handleChange}
+              onChange={(event) =>
```
sebastien-comeau, line 202: Create a handle callback function:

```tsx
const handleOnESRFChange: ChangeEventHandler<HTMLInputElement> = useCallback(({ target }) => {
  formik.setFieldValue(target.name, target.value.replace(/[^a-z0-9]/gi, ''))
}, [formik])
```

public/locales/en/status.json
krischarbonneau: I'm okay with this given the recent introduction of letters into the ESRF, but given the original `length` string was provided to us by the BAs this might change.

public/locales/fr/status.json
krischarbonneau: I'm okay with this given the recent introduction of letters into the ESRF, but given the original `length` string was provided to us by the BAs this might change.

pages/status.tsx
krischarbonneau: Had a discussion about useCallback with Stefan since I originally didn't see the need to add a handler outside of the `onChange` that used the same function, but given some examples I've looked at and the docs I think this makes sense given we're using Formik.

sebastien-comeau (approved):
Looks good to me. 👍

PR #127, opened 2022-11-15
Refactor MetaData component

sebastien-comeau:
### Description

List of proposed changes:

- MetaData component refactoring.
- Removed keywords prop because search engine like Google does not use the keywords meta tag in web ranking. https://developers.google.com/search/blog/2009/09/google-does-not-use-keywords-meta-tag

Load the application and make sure page header tags are set on every pages.


krischarbonneau (approved):
Looks good, was wondering if we were going to get rid of keywords at some point since we weren't even using it

PR #128, opened 2022-11-15
Fix given names

Maxim-Lam:
### Description

List of proposed changes:

changed given name -> given name(s)

go to /status and /email pages

### DIFF
--- a/public/locales/en/email.json
+++ b/public/locales/en/email.json
@@ -10,7 +10,7 @@
   },
   "given-name": {
     "error": {
-      "required": "The given name is required."
+      "required": "The given name(s) is required."
     },
     "label": "Given name(s)"
   },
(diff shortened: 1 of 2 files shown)


krischarbonneau (approved):

krischarbonneau (changes requested):
As mentioned over Slack, we'll need to change the French translation for this as well, and adjust the error messaging so it's not singular. Willy rely on @sebastien-comeau for proper French translations but my understanding after asking a couple people was that:

`The given name(s) is required` should be something along the lines of `Given name(s) required` and the French should be `Prénom(s) obligatoire`

sebastien-comeau (approved):

krischarbonneau (approved):

PR #129, opened 2022-11-17
Refactor API endpoints

sebastien-comeau:
### Description

List of proposed changes:

- Refactor API endpoints to have similar code logic. Move away from throwing ApiError excpetion and returning fetch Response object
- Rename and move TS types
- Reorganize `lib` directory

Load the application and test both forms.


krischarbonneau (approved):
Looks good. Like that you moved `compareCIAndAI` into utils

PR #133, opened 2022-11-23
Remove prop-types dependency in favor of TS

sebastien-comeau:
### Description

List of proposed changes:

- Remove `prop-types` dependency because we moved on with TypeScript
-


krischarbonneau (approved):
Looks good

PR #135, opened 2022-11-23
Added 4 digit max input for date field. Only works on chromium

Maxim-Lam:
### Description

added 4 digit limit on date field. Only works on chromium browsers.
List of proposed changes:

added the 4 digit limit on date field

go to /email and /status

try entering more than 4 digits in year


sebastien-comeau (approved):
Looks good!

PR #136, opened 2022-11-23
Add idle timeout to form pages

krischarbonneau:
### Description
This PR leverages the [React Idle Timer](https://idletimer.dev/) library to prompt a user after 15 minutes of idle time (based on mouse & keyboard events) if they would like to be redirected or not. I modified the handlers for the modal on both `/email` and `/status` so that the modal can either open when the user selects the cancel button or when the timeout limit of 15 minutes is reached. If the modal opens as a result of a timeout and the user selects `yes`, the `agreed-to-email-esrf-terms` cookie is reset so that they are shown the `/expectations` page again (in the case of the site being accessed on a public computer), else the user is redirected to `/landing` like normal.

1. Pull in branch
2. Type `npm run dev`
3. Agree to terms and then navigate to `/status`
4. Set the value of `timeout` to `3000` on line 67
5. If you don't move your mouse for 3 seconds, the modal should pop up indicating that you've been inactive for 15 minutes.
6. Selecting `No` will close the modal (and since the timeout is set to 3000ms in this case, it will pop up again after 3 seconds of inactivity). Selecting yes will remove the `agreed-to-email-esrf-terms` cookie and redirect you to `/expectations`.
7. Confirm the modal stills works as expected when clicking the `Cancel` button normally.
8. Repeat steps for `/email` page

### DIFF
--- a/pages/email.tsx
+++ b/pages/email.tsx
@@ -16,6 +16,8 @@ import Modal from '../components/Modal'
 import LinkSummary, { LinkSummaryItem } from '../components/LinkSummary'
 import useEmailEsrf from '../lib/useEmailEsrf'
 import { EmailEsrfApiRequestBody } from '../lib/types'
+import { useIdleTimer } from 'react-idle-timer'
+import { setCookie } from 'cookies-next'
 
 const initialValues: EmailEsrfApiRequestBody = {
   dateOfBirth: '',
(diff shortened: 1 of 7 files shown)


pages/email.tsx
```diff
@@ -27,6 +29,35 @@ const initialValues: EmailEsrfApiRequestBody = {
 export default function Email() {
   const { t } = useTranslation('email')
   const [modalOpen, setModalOpen] = useState(false)
```
sebastien-comeau, line 31: Here's my recommended code refactoring:

```tsx
  const router = useRouter()
  const [modalOpen, setModalOpen] = useState(false)
  const [isIdle, setIsIdle] = useState(false)

  const handleOnIdleTimerIdle = useCallback(() => {
    setIsIdle(true)
    setModalOpen(true)
  }, [])

  const { reset: resetIdleTimer } = useIdleTimer({
    onIdle: handleOnIdleTimerIdle,
    //15 minute timeout
    timeout: 15 * 60 * 1000,
  })

  const handleOnModalRedirectButtonClick = useCallback(() => {
    //If user is idle and selects option to go back, clear the cookie so they get redirected to /expectations instead
    if (isIdle) {
      deleteCookie('agreed-to-email-esrf-terms')
    }
    router.push('/landing')
  }, [isIdle, router])

  const handleOnModalResetButtonClick = useCallback(() => {
    setModalOpen(false)
    if (isIdle) {
      setIsIdle(false)
      resetIdleTimer()
    }
  }, [isIdle, resetIdleTimer])
```

The main changes are:

- Use `useRouter` hook instead of `Router`
- Change the timeout calculation from 150 * 6 * 1000` to `15 * 60 * 1000`
- Use `useCallback` hook
- Rename handler method name like `handle{EventThingNameHere}`
- Use `deleteCookie` instead of `setCookie`

sebastien-comeau (commented):
Some recommended code refactoring.

pages/email.tsx
sebastien-comeau: This can also be applied to `email.tsx` page.

sebastien-comeau (approved):
We can have a discussion on how this timeout idle modal should work. I thought after 15 mins of inactivity, the user would be forced to start over otherwise someone else would be able ton continue with the page state if the user forgot to close the page. We can have a chat on this but other than that it's good.

PR #139, opened 2022-11-28
Remove trigger button from Modal

sebastien-comeau:
### Description

List of proposed changes:

- Remove trigger button from Modal component
- Add button disabled styling

Pages status, email and contact should work as attented.

### Additional Notes

The main motivation behind this it's because the trigger button should be inside the Modal because a modal can be triggered by other events like in our case the user idle event.

I've added a cypress tests when clicking cancel button but at the moment the a11y test is skipped because the modal doesn't pass a11y.
image


krischarbonneau (approved):
Looks good to me, can figure out the testing seperately.

PR #148, opened 2022-12-05
Fix email page style

sebastien-comeau:
### Description

List of proposed changes:

- Fix email page style
- Fix phone-number i18n key

Load email page and fill the form. Take a look at the style and phone number.


krischarbonneau (approved):
Looks good!

PR #149, opened 2022-12-05
Fix input label a11y

sebastien-comeau:
### Description

List of proposed changes:

- Fix input label by adding aria-label attribute

Load the form and make sure the inputs have a aria-label attribute pointing to the correct label id.


krischarbonneau (approved):
Looks good!

PR #151, opened 2022-12-05
Fix check-status no record component style

sebastien-comeau:
### Description

List of proposed changes:

- Fix check-status no record style

Load the application, go to check-status page, enter invalid  data and check the no record component style.


krischarbonneau (approved):
Looks good!

PR #152, opened 2022-12-05
Fix 404 and error pages a11y

sebastien-comeau:
Use heading classes instead of element css style

### Description

List of proposed changes:

- Fix 404 and error page a11y
- Change element css style with class style for headings

I've created a cypress test to detects a11y issues for 404 page. That part should be covered.

### Additional Notes

I needed to move away from css element style to class style for headings since the 404 page requires `h1` style on `h2` html tag. It's a common pattern with MUI framework.


krischarbonneau (approved):
Looks good!

PR #159, opened 2022-12-06
Fix localisations and validation errors order

sebastien-comeau:
### Description

List of proposed changes:

- Fix some localisations
- Fix validation error messages order

Load the application and make sure the error messages order is the same as the fields order.


krischarbonneau (approved):
Looks good!

PR #165, opened 2022-12-06
Added scroll to top functionality for status page

krischarbonneau:
### Description
Currently, if you click `Check Status` on the `/status` page the status information is conditionally rendered. The problem here is that since we aren't actually routing to a new page, our scroll position is maintained when showing the status info. This isn't very noticeable on desktop resolutions but it is very noticeable on mobile resolutions.

My solution was to add a function that scrolls to the top of the page and call this function when the `Previous` button is clicked and before `CheckStatusInfo` is rendered.

1. Pull in branch
2. Type `npm run dev`
3. In developer tools, set your viewport to a mobile viewport (I used iPhone XR for testing purposes)
4. Navigate to `/status`
5. Fill in form field with dummy data so that `<CheckStatusNoRecord/>` is rendered and ensure your scroll position is at the top of the page
6. Pressing previous should send you back to the pre-filled form with your scroll position at the top of the page

### DIFF
--- a/pages/status.tsx
+++ b/pages/status.tsx
@@ -5,7 +5,6 @@ import {
   useCallback,
   useMemo,
   useState,
-  useEffect,
 } from 'react'
 import { GetStaticProps } from 'next'
 import { useRouter } from 'next/router'
--- a/pages/status.tsx
+++ b/pages/status.tsx
@@ -5,6 +5,7 @@ import {
   useCallback,
   useMemo,
   useState,
+  useEffect,
 } from 'react'
 import { GetStaticProps } from 'next'
 import { useRouter } from 'next/router'
@@ -125,7 +126,6 @@ const Status: FC = () => {
       }
       formik.setStatus(undefined)
       removeCheckStatusResponse()
-      scrollToTop()
     },
     [formik, removeCheckStatusResponse, checkStatusResponse, router]
   )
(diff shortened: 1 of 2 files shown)


pages/status.tsx
```diff
@@ -136,6 +137,11 @@ const Status: FC = () => {
     [formik]
   )
 
+  //When the page is conditionally rendered it maintains its spot on the page. Calling this will force it to scroll to the top of the page.
+  const scrollToTop = () => {
```
sebastien-comeau, line 141: Use `useEffect` hook instead.

```
useEffect(() => {
  window.scrollTo(0, 0)
}, [checkStatusResponse])
```

Page will scroll to top on component mount and when his dependency `checkStatusResponse` change.

pages/status.tsx
```diff
@@ -148,6 +154,7 @@ const Status: FC = () => {
       <h1 className="h1">{t('header')}</h1>
       {(() => {
         if (checkStatusResponse !== undefined) {
+          scrollToTop()
```
sebastien-comeau, line 157: This will make the page scroll everytime the component re-render. Not recommended. If you want to scroll to top on event then you can add event handler on `useCheckStatus` options.

pages/status.tsx
sebastien-comeau: https://stackoverflow.com/a/39345165

krischarbonneau: I think this is the way I had it originally but without the dependency. Settled on the function instead but this is much cleaner, thanks.

PR #167, opened 2022-12-07
Add error summary scroll and focus on field errors

sebastien-comeau:
### Description

List of proposed changes:

- Add error summary scrolling and focusing on form validation errors.

Load the application and navigate and submit forms. The error summary should be scrolled into view and focus to help with keyboard navigation.

### DIFF
--- a/__tests__/components/ErrorSummary.test.tsx
+++ b/__tests__/components/ErrorSummary.test.tsx
@@ -6,6 +6,11 @@ import ErrorSummary from '../../components/ErrorSummary'
 expect.extend(toHaveNoViolations)
 
 describe('ErrorSummary', () => {
+  /**
+   * @see https://stackoverflow.com/a/60225417
+   */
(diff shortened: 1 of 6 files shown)


Maxim-Lam:
@sebastien-comeau is there a way that you can add a little margin or padding to the top of the scroll? When clicking the file number error it brings me straight to the field but cuts off the label. Just a little confusing.
image

krischarbonneau:
> > @sebastien-comeau is there a way that you can add a little margin or padding to the top of the scroll? When clicking the file number error it brings me straight to the field but cuts off the label. Just a little confusing.
> 
> @Maxim-Lam Can't change this browser behavior because it's a anchor link to the input id.
> 
> WET template do the same: https://wet-boew.github.io/v4.0-ci/demos/formvalid/formvalid-server-en.html

I agree with Max that it isn't the best user experience, but considering that it is the same functionality as WET, I'm okay with this .

krischarbonneau (approved):

PR #168, opened 2022-12-07
Fix scroll to top so it no longer utilizes useEffect

krischarbonneau:
### Description

List of proposed changes:

- Remove usage of `useEffect` and instead scroll to top when `useCheckStatus` is successful or when clicking the `Previous` button

1. Pull in branch
2. Type `npm run dev`
3. Navigate to `/status`
4. View the page on a mobile resolution
5. Enter any information so `<CheckStatusNoRecord/>` is rendered and confirm that you are sent to the top of the screen
6. Ensure when clicking `Previous` that you are sent to the top of the screen


sebastien-comeau (approved):
Looks good!

PR #169, opened 2022-12-08
Add and integrate reusable IdleTimout component

sebastien-comeau:
### Description

List of proposed changes:

- Move idle timout code duplication into IdleTimeout reusable component

- You can modify the timeout and promptTimeout IdleTimer options with small values (in milliseconds) when testing. The idle timeout modal should appears like usual. If a modal is already shown then it will be hidden temporarily then if use continue the session the modal will show up again.

### DIFF
--- a/components/IdleTimeout.tsx
+++ b/components/IdleTimeout.tsx
@@ -1,0 +1,74 @@
+import { deleteCookie } from 'cookies-next'
+import { useTranslation } from 'next-i18next'
+import { useRouter } from 'next/router'
+import { FC, useState } from 'react'
+import { IIdleTimerProps, useIdleTimer } from 'react-idle-timer'
+import Modal from './Modal'
+
+export interface IdleTimeoutProps
+  extends Pick<IIdleTimerProps, 'promptTimeout'>,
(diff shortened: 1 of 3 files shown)


components/IdleTimout.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { useRouter } from 'next/router'
+import { FC, useState } from 'react'
+import { IIdleTimerProps, useIdleTimer } from 'react-idle-timer'
+import Modal from './Modal'
+
+export interface IdleTimoutProps
+  extends Pick<IIdleTimerProps, 'promptTimeout'>,
+    Pick<IIdleTimerProps, 'timeout'> {}
+
+const IdleTimout: FC<IdleTimoutProps> = ({ promptTimeout, timeout }) => {
```
krischarbonneau, line 12: Should read IdleTimeout

krischarbonneau (commented):
Works well, just a slight spelling error

components/IdleTimout.tsx
sebastien-comeau: Good catch :), fixed.

krischarbonneau (approved):

PR #171, opened 2022-12-09
Tweak index page style to match Canada.ca

sebastien-comeau:
### Description

List of proposed changes:

- Tweak index page style to match Canada.ca

Compare index page style with Canada.ca language box.


krischarbonneau (approved):
Looks good!

PR #172, opened 2022-12-09
Fix layout Canada logo style

sebastien-comeau:
### Description

List of proposed changes:

- Tweak Canada logo style in layout header to match Canada.ca


krischarbonneau (approved):
Looks good 👍

PR #174, opened 2022-12-09
Fix anchor pseudo css class order

sebastien-comeau:
### Description

List of proposed changes:

- Fix anchor pseudo css class order


krischarbonneau (approved):
Works well!

PR #175, opened 2022-12-09
Added file number images to status page

Maxim-Lam:
### Description

List of proposed changes:

Added images under an expand/collapse component. 

I think Design systems describes collapse widgets use the best:
image

navigate to /status

toggle collapse in both french and english


components/Collapse.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export interface CollapseProps {
+  title: string
+  children?: JSX.Element
+}
+
+const Collapse: FC<CollapseProps> = ({ title, children }) => {
+  return (
+    <details className="max-w-prose border p-3 mb-3 rounded">
+      <summary className="text-blue-light">{title}</summary>
```
krischarbonneau, line 11:
```suggestion
      <summary className="text-blue-light hover:underline hover:text-link-selected cursor-pointer focus:text-link-selected focus:underline">{title}</summary>
```

Following the design seen [here](https://design.canada.ca/common-design-patterns/collapsible-content.html), there should be hover and focus styling on the summary element

krischarbonneau (changes requested):

components/Collapse.tsx
```diff
@@ -0,0 +1,19 @@
+import { FC } from 'react'
+
+export interface CollapseProps {
+  title: string
+  children?: JSX.Element
```
sebastien-comeau, line 5: Should be the following type:

```tsx
children?: React.ReactNode
```

sebastien-comeau (commented):
Minor tweak and to fix the failing status page tests you need to mock ExampleImage component.
```js
jest.mock('../../components/ExampleImage')
```

sebastien-comeau:
My only concern is accessibility. I wonder if the details or summary element should have an id and the input have this id in his aria-describedby attribute. I've searched for this information but didn't find any good info with details/summary element.

Maxim-Lam:
> My only concern is accessibility. I wonder if the details or summary element should have an id and the input have this id in his aria-describedby attribute. I've searched for this information but didn't find any good info with details/summary element.

I have added an aria-describeby as you suggested. Let me know if there is anything else

krischarbonneau (approved):
Looks good!

PR #176, opened 2022-12-12
update modal styling and add countdown timer

krischarbonneau:
### Description

This PR updates the modal styling to be more in line with other Canada.ca design patterns for modals and also updates the idle description to include a countdown timer to let the user know how much time they have before they are redirected.

1. Pull in branch
2. Type `npm run dev`
3. Change the value for `timeout` to something small
4. The modal should pop up warning of a session timeout, with a timer counting down from 5 minutes (or whatever you set `promptTimeout` to)

### DIFF
--- a/components/IdleTimeout.tsx
+++ b/components/IdleTimeout.tsx
@@ -48,7 +48,7 @@ const IdleTimeout: FC<IdleTimeoutProps> = ({ promptTimeout, timeout }) => {
     onIdle: handleOnIdle,
     onPrompt: handleOnPrompt,
     promptTimeout: promptTimeout ?? 5 * 60 * 1000, //5 minutes
-    timeout: timeout ?? 10, //10 minutes
+    timeout: timeout ?? 10 * 60 * 1000, //10 minutes
   })
 
   const tick = useCallback(() => {
@@ -70,17 +70,17 @@ const IdleTimeout: FC<IdleTimeoutProps> = ({ promptTimeout, timeout }) => {
         {
           onClick: () => handleOnIdle(),
           style: 'primary',
-          text: t('modal.idle-end-session'),
+          text: t('modal-idle-timeout.end-session'),
         },
         {
           onClick: handleOnIdleContinueSession,
           style: 'default',
-          text: t('modal.idle-continue-session'),
+          text: t('modal-idle-timeout.continue-session'),
         },
       ]}
-      header={t('modal.idle-header')}
+      header={t('modal-idle-timeout.header')}
     >
-      <p>{t('modal.idle-description', { timeRemaining })}</p>
+      <p>{t('modal-idle-timeout.description', { timeRemaining })}</p>
     </Modal>
   )
 }
--- a/components/IdleTimeout.tsx
+++ b/components/IdleTimeout.tsx
@@ -1,7 +1,7 @@
 import { deleteCookie } from 'cookies-next'
 import { useTranslation } from 'next-i18next'
 import { useRouter } from 'next/router'
-import { FC, useState } from 'react'
+import { FC, useState, useEffect, useCallback } from 'react'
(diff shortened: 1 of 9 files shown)


components/IdleTimeout.tsx
```diff
@@ -43,14 +44,26 @@ const IdleTimeout: FC<IdleTimeoutProps> = ({ promptTimeout, timeout }) => {
...
     onIdle: handleOnIdle,
     onPrompt: handleOnPrompt,
-    promptTimeout: promptTimeout ?? 5 + 60 * 1000, //5 minutes
+    promptTimeout: promptTimeout ?? 5 * 60 * 1000, //5 minutes
     timeout: timeout ?? 10 * 60 * 1000, //10 minutes
   })
 
-  return (
+  const tick = useCallback(() => {
+    var minutes = Math.floor(getRemainingTime() / 60000)
```
sebastien-comeau, line 55: `var` should be `const`

components/IdleTimeout.tsx
```diff
@@ -43,14 +44,26 @@ const IdleTimeout: FC<IdleTimeoutProps> = ({ promptTimeout, timeout }) => {
...
+    var seconds = Math.floor((getRemainingTime() / 1000) % 60).toFixed(0)
```
sebastien-comeau, line 56: `var` should be `const`

components/IdleTimeout.tsx
```diff
@@ -43,14 +44,26 @@ const IdleTimeout: FC<IdleTimeoutProps> = ({ promptTimeout, timeout }) => {
...
+    setTimeRemaining(
+      minutes + ':' + (parseInt(seconds) < 10 ? '0' : '') + seconds
+    )
+  }, [getRemainingTime])
+
+  useEffect(() => {
+    setInterval(tick, 1000)
+  }, [tick])
+
+  return timeRemaining ? (
```
sebastien-comeau, line 66: This could be checked at the Model.open prop.

```tsx
open={modalOpen && timeRemaining.lenght > 0}
```

components/Modal.tsx
```diff
@@ -19,13 +25,19 @@ const Modal: FC<ModalProps> = ({ actionButtons, children, open }) => {
       >
         <div
           role="dialog"
-          aria-describedby={`${id}-modal-desc`}
-          className="mx-4 p-4 bg-white border-2 border-black md:w-2/3 lg:w-2/5"
+          aria-describedby={`${id}-modal-header`}
+          className="bg-white ring-2 ring-gray-modal md:w-2/3 lg:w-2/5 rounded"
```
sebastien-comeau, line 29: `mx-4` class is important for mobile view... othewise the model is full width... should have x-axis margin

components/Modal.tsx
```diff
@@ -1,14 +1,20 @@
...
+import { FC, useId } from 'react'
 import ActionButton, { ActionButtonProps } from './ActionButton'
 import { FocusOn } from 'react-focus-on'
 
 export interface ModalProps {
   actionButtons: ActionButtonProps[]
-  children: ReactNode
   open: boolean
+  header: string
+  description: string
```
sebastien-comeau, line 9: I would still use `children: ReactNode` instead of description. It gives you more flexibility for the model content, as example if you want to display multiple paragraph.

public/locales/en/common.json
```diff
@@ -61,8 +61,10 @@
     "cancel-button": "Cancel",
```
sebastien-comeau, line 61: These keys should be grouped by their modal concerns. Like "modal-go-back" and "modal-idle-timeout".

components/Modal.tsx
krischarbonneau: good catch on this 😄

sebastien-comeau (approved):
Approve!!! Good job!

PR #179, opened 2022-12-13
Remove LinkSummary component and contact page

krischarbonneau:
### Description
This PR removes the `LinkSummary` component completely as we do not need it anymore per this morning's standup. I've also removed the contact page since it is no longer part of our flow. All translations relating to both the component and page have been removed as well.

1. Pull in branch
2. Type `npm run dev`
3. Ensure that navigating to `/contact` results in a 404
4. Navigate to `/status` and fill in the form with random info so the No Record status is shown and confirm that the contact links are no longer at the bottom


Maxim-Lam (approved):
LGTM

PR #184, opened 2022-12-15
Add preferred language for email file number

sebastien-comeau:
### Description

List of proposed changes:

- Add preferred language name in email file number payload based on UI locale


krischarbonneau (approved):
Looks good!

PR #185, opened 2022-12-15
Added feedback link to /status and /email pages using i18next Trans c…

Maxim-Lam:
…omponent.

### Description

List of proposed changes:

- Added feedback link text and placeholder link to /email and /status result pages
- easier way to add full translated sentences with translated links in the middle of them to page. We should probably implement where we can

- go to /email
- enter dummy valid data everywhere
- notice text under button
- repeat for /status

### Additional Notes

Link url to come

### DIFF
--- a/pages/status.tsx
+++ b/pages/status.tsx
@@ -148,8 +148,8 @@ const Status: FC = () => {
                 checkStatusResponse={checkStatusResponse}
               />
               <div className="mt-10">
-                <Trans i18nKey={'feedback-link'}>
-                  Insert feedback <a href="">Link</a>
+                <Trans i18nKey={'common:feedback-link'}>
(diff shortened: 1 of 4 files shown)


pages/status.tsx
```diff
@@ -146,6 +147,11 @@ const Status: FC = () => {
                 checkAgainText={t('check-again')}
                 checkStatusResponse={checkStatusResponse}
               />
+              <div className="mt-10">
+                <Trans i18nKey={'feedback-link'}>
```
sebastien-comeau, line 151: Add `common` namespace to the key as `'common:feedback-link'`. Apply it in email page as well.

sebastien-comeau (approved):

PR #186, opened 2022-12-16
Refactor Layout component

sebastien-comeau:
### Description

List of proposed changes:

- Refactor Layout component and use `useTranslation` hook within the component.
- Move away of using `returnObjects: true` with `useTranslation` hook because error can only be detected at runtime.
- Change page meta title to `Check the status of your passport application - Canada.ca` in English and `Vérifiez l'état de votre demande de passeport - Canada.ca` in French. Official page meta title and description will be provided.
- Fix footer privacy-url and terms-and-condition-url in french local.

Load the application and ensure the layout locales are correct.


krischarbonneau (approved):
Looks good!

PR #187, opened 2022-12-16
Cleanup with TS config changes

sebastien-comeau:
### Description

List of proposed changes:

- Removed unused variables or imports
- Set TS config "noImplicitAny" and "noUnusedLocals" to true

Load the application and run `npm run build` and shouldn't not have any TS errors

### Additional Notes
https://www.typescriptlang.org/tsconfig#noImplicitAny
https://www.typescriptlang.org/tsconfig#noUnusedLocals

### DIFF
--- a/components/Footer.tsx
+++ b/components/Footer.tsx
@@ -1,5 +1,4 @@
 import { FC } from 'react'
-import { useTranslation } from 'next-i18next'
 import DateModified from './DateModified'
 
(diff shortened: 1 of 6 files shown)


krischarbonneau (approved):
Looks good. `noUnusedLocals` is a good add since unused imports have been left in a few times now so that will help to prevent that.

PR #191, opened 2022-12-19
External link a11y

krischarbonneau:
### Description
This PR adds the `ExternalLink` component which is just an enhanced `a` tag that announces to the screen reader that the link opens in a new tab. Global CSS was added that appends an external link icon to the end of any `a` tag that includes `target="_blank"`.

1. Pull in branch
2. Type `npm run dev`
3. Navigate to `/status`
4. Enter in the following mock data:
```
File #: A081EFD7
Given Name: Nicolas
Surname: Roux
DOB: 1959-12-06
```
5. The above data should show `Shipping via Canada Post` and the link next to the tracking number should have an external icon and open in a new tab.


sebastien-comeau (approved):
Looks good!

PR #196, opened 2022-12-22
Add Test site text on non-prod environments

krischarbonneau:
### Description
This PR adds the `Banner` component which is only displayed if the environment variable `NEXT_PUBLIC_ENVIRONMENT` is anything other than `prod`

1. Pull in branch
2. Add the line `NEXT_PUBLIC_ENVIRONMENT=dev` to your `.env.local`
3. Type `npm run dev`
4. Ensure that the `Banner` component is rendered on `/expectations`
5. Change the value of the environment variable to `prod` and ensure that the component no longer renders


components/Banner.tsx
```diff
@@ -0,0 +1,30 @@
+import { FC } from 'react'
+
+export interface BannerProps {
+  alert: string
+  description: string
+}
+
+const Banner: FC<BannerProps> = ({ alert, description }) => {
+  return (
+    <div className="bg-blue-normal">
```
sebastien-comeau, line 10: Change the html and css to the following which is less confusing and remove a lot of uneeded css classes.

```tsx
<div className="bg-blue-normal font-body text-white">
      <div className="container mx-auto p-4 flex flex-col space-y-2 lg:flex-row lg:items-center lg:space-x-4 lg:space-y-0">
        <div
          className="border-2 px-4 py-1 whitespace-nowrap w-max"
          role="alert"
        >
          <b>{alert}</b>
        </div>
        <div>{description}</div>
      </div>
    </div>
```

sebastien-comeau (commented):
Small html/css tweak.

Here's a good guide to Flexbox that I'm always refering to when needed:
https://css-tricks.com/snippets/css/a-guide-to-flexbox/

components/Header.tsx
```diff
@@ -32,6 +35,12 @@ const Header: FC<HeaderProps> = ({ gocLink, skipToMainText }) => {
       </nav>
 
       <header>
+        {process.env.NEXT_PUBLIC_ENVIRONMENT !== 'prod' && (
```
sebastien-comeau, line 38: Add this env. variable to `.env.example` file for reference.

components/Banner.tsx
krischarbonneau: Was going off an existing solution but this is much cleaner, thanks

sebastien-comeau (approved):

PR #197, opened 2022-12-22
Add Date Select Field component

sebastien-comeau:
### Description

List of proposed changes:

- Created and implemented a DateSelectField component

Run the application and fill the form. The input date should be replcae with the DateSelectField that have 3 selects (year, month and day). If you select a wrong date, like Feburary 29 in a common year (non leap year) then the day should be cleared forcing the user to choose the day again.

### Additional Notes

Source of inspiration
https://github.com/whitphx/react-ymd-date-select/blob/main/src/use-date-select.ts


Maxim-Lam (changes requested):
I'm wondering if it would be better to have the input field span the length of the screen on mobile devices. As it stands, it's about a third of the screen; the examples that stef linked to us had the input the width of the screen.

image

sebastien-comeau:
> I'm wondering if it would be better to have the input field span the length of the screen on mobile devices. As it stands, it's about a third of the screen; the examples that stef linked to us had the input the width of the screen.

I think it looks weird full width when you have max 4 digits values.

Maxim-Lam (approved):
LG

PR #207, opened 2023-01-04
Update status fields error message

sebastien-comeau:
### Description

List of proposed changes:

- Update fields error message to match new field's name

Load the `status` page and rendre the error messages. The field name in them should match the field's name.


krischarbonneau (approved):
Looks good!

PR #216, opened 2023-01-04
Bump next-i18next to 13.0.3

krischarbonneau:
### Description
This PR fixes the issue causing #200's tests to fail. I've created a new PR rather than updating that one since there's been a new minor version released.

More information on why this error occurs and the official fix for it per i18next docs can be seen [here](https://www.i18next.com/overview/typescript#argument-of-type-defaulttfuncreturn-is-not-assignable-to-parameter-of-type-xyz)

1. Pull in branch
2. Type `npm run tsc:validate` and ensure there are no failures
3. Type `npm run dev` and ensure the application works as expected


sebastien-comeau (approved):
Good job!

PR #217, opened 2023-01-05
Bump some packages version

sebastien-comeau:
### Description

List of proposed changes:

- Bump some packages version

Load the branch code, run `npm install`, run tests `npm run test`, run app build `npm run build` and browse app `npm run start`.

### Additional Notes

Some Cypress tests was falling when tesitng for a11y issues. After reading, newer version of cypress will isolate each test and start each of them with a blank page. If you load a page state for one test then doing another test on this state thinking the page will be the same then you are wrong. 

> Best Practice: Tests should always be able to be run independently from one another and still pass .

https://docs.cypress.io/guides/core-concepts/test-isolation


krischarbonneau (approved):
Meaty PR! Looks good 😄

PR #219, opened 2023-01-06
Next SEO integration

sebastien-comeau:
### Description

List of proposed changes:

- Next SEO Integration

Load the branch code locally and run the application `npm run dev`. These's should have header meta tag to support multiple things like Facebook (open graph protocol), Twitter and DCTerms. The index page robot instruction should be `noindex,follow` but `index,follow` for the other pages by default.

If you add the env. variable `NEXT_PUBLIC_APP_BASE_URI=http://localhost:3000` in your `.env.local` file then you shoul see the language alternate links in the header and also the open graph image tags.

### Additional Notes
Next SEO
https://github.com/garmeeh/next-seo


krischarbonneau (approved):
This is great!

PR #223, opened 2023-01-09
Use publicRuntimeConfig instead of NEXT_PUBLIC_

sebastien-comeau:
### Description

List of proposed changes:

- Use publicRuntimeConfig instead of NEXT_PUBLIC_ environment variables in Pages
- Use server-side rendering in all pages

Load the branch code locally, run `npm run build` and `npm run start` and load the application. You shouldn't see the following tags in head tag.

```html
<link rel="alternate" hreflang="en" href="http://localhost:3000/en/expectations">
<link rel="alternate" hreflang="fr" href="http://localhost:3000/fr/expectations">
<meta property="og:image" content="http://localhost:3000/ogp.jpg">
<meta property="og:image:width" content="2048">
<meta property="og:image:height" content="1152">
```

Set the following env. variable in your `.env.local` file then stop the process and start the application `npm run start` **WITHOUT** building the application again. At this point you should see the above html tags in head tag.

```yaml
# Base URI for the Application
NEXT_PUBLIC_APP_BASE_URI=http://localhost:3000
```

### Additional Notes

Using `process.en.NEXT_PUBLIC_APP_BASE_URI` directly is not good because the value is baked at build time. Setting the `NEXT_PUBLIC_*` variables value at runtime doesn't applied them. The application needs to leverage on using runtime configuration with `publicRuntimeConfig` and switching all our pages to server-side rendering method.
https://nextjs.org/docs/api-reference/next.config.js/runtime-configuration

How to keep NextJs configurable in Docker
https://www.witodelnat.eu/blog/2021/nextjs-docker

### DIFF
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -3,6 +3,7 @@ import Link from 'next/link'
 import { useRouter } from 'next/router'
 import Banner from './Banner'
 import { useTranslation } from 'next-i18next'
+import getConfig from 'next/config'
 
 export interface HeaderProps {
   gocLink: string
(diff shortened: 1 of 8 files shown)


krischarbonneau (approved):
Looks good!

PR #227, opened 2023-01-09
Temporarily update titles for each page to pass a11y

krischarbonneau:
### Description
The only remaining item from our a11y audit (that is a failure, there are two suggestions) is that all of our pages have the same title. Given that we are waiting for comms to provide us with the metadata, this is a temporary fix. I've hardcoded the text rather than using translations to switch between `en | fr` and `fr | en` given the fact that this is all likely to change in the next few days. The only purpose of this PR is to get the app to pass the a11y audit. A more permanent solution (ie. actually editing the Next SEO config) will be done when we get the content from comms

1. Pull in branch
2. Type `npm run dev`
3. Navigate through each page and confirm that the page title is unique

### DIFF
--- a/pages/email.tsx
+++ b/pages/email.tsx
@@ -21,6 +21,7 @@ import ExternalLink from '../components/ExternalLink'
 import DateSelectField, {
   DateSelectFieldOnChangeEvent,
 } from '../components/DateSelectField'
+import { NextSeo } from 'next-seo'
 
 const initialValues: EmailEsrfApiRequestBody = {
   dateOfBirth: '',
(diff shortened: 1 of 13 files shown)


sebastien-comeau (changes requested):
No need to set `titleTemplate` because it is set at application level in `_app.tsx` using `DefaultSeo` component. The default seo content is in `next-seo.config.ts`. Set the `titlte` prop by using the `useTranslation` hook. We can update the English and French title later by updating the `JSON` locales.

sebastien-comeau (approved):
Looks good! :)

PR #231, opened 2023-01-11
Fix date select error navigation

sebastien-comeau:
### Description

List of proposed changes:

- Changed the DateSelect component to be able to navigate to the first select (year) when clicking on the error message link.

Load the application and run it. Generate an error message for the date of birth field and click on the error message at the top. The date of birth year select should be focused.


Maxim-Lam (approved):
LGTM!

PR #247, opened 2023-01-20
Handle HTTP422 responses from API as `not found`

gregory-j-baker:
### Description

List of changes:

- Fix regression from commit `b0208241` that broke how 422 response codes from the API are handled.

Perform a status search that will result in multiple search results, expect to see a 404 from the API instead of a 422.


sebastien-comeau (approved):
Big bag of chips for you! :)

PR #248, opened 2023-01-20
Change application name and meta data

sebastien-comeau:
### Description

List of proposed changes:

- Make the page title to reflect of page h1 or application
- Change the application name in English `Passport application status checker` and French `Outil de vérification de l'état de la demande de passeport`
- Change `statut` to `état` in French content

Ensure the application title is correct in both English and French.

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -8,7 +8,7 @@ const Index = () => {
     <>
       <NextSeo
         noindex
-        title="Passport Application Status Checker (PASC) | Vérificateur du Statut de mon application pour un passport (VSAP)"
+        title="Passport application status checker | Outil de vérification de l'état de la demande de passeport"
         titleTemplate={'%s \u2010 Canada.ca'}
       />
       <main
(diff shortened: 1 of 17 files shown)


pages/index.tsx
```diff
@@ -18,7 +18,8 @@ const Index = () => {
         <div className="m-auto w-[300px] md:w-[400px] lg:w-[500px] bg-gray-lighter">
           <div className="p-8">
             <h1 className="sr-only">
-              service.canada.ca-passport-status-checker
+              Passport application status checker | Outil de vérification de
+              l&apos;état de la demande de passeport
```
gregory-j-baker, line 22: `&apos;` should be changed to `'` to be consistent with the rest of the application.

gregory-j-baker (changes requested):

pages/index.tsx
sebastien-comeau: We are actually using `&#39;` in other pages. Eslint will complains if we don't escape `'` in the html.
image

sebastien-comeau: I've updated that file with `&#39;`

sebastien-comeau: Because it's a bilingual page and not translated.

gregory-j-baker (approved):

PR #267, opened 2023-01-24
Use Trans component in status components

sebastien-comeau:
### Description

List of proposed changes:

- Use Trans component in status components.

Load status `1` and `no-record` found compoments.


ken-blanchard (approved):
Looks good.

PR #269, opened 2023-01-25
Fix expectations content case

sebastien-comeau:
### Description

List of proposed changes:

- Fix expectations content case in French


Maxim-Lam (approved):
Excellent work here

PR #283, opened 2023-01-27
Removed fr reference to sms

Maxim-Lam:
### Description

List of proposed changes:

Remove the term SMS from FR page


sebastien-comeau (approved):
Looks good!

PR #288, opened 2023-01-30
Update text to use interpolation, use strong instead of b

krischarbonneau:
### Description
This PR accomplishes a few things. I've updated `CheckStatusNoRecord` to use interpolation instead of using the `or` key (and have removed that key since this was the last instance of it being used). I've switched all instances where the `<b>` tag was being used to `<strong>` as per the a11y audit (excluding the banner since that won't be on prod anyway). I placed the `We update the status of applications daily` text inside a `<p>` tag as per the a11y audit.

Aside from the above, I've also done some general clean-up on our translation files to get rid of keys that were no longer being used.

1. Pull in branch
2. Type `npm run dev`
3. Navigate to `/expectations`, inspect page and ensure that the text indicated above is in a `<p>` tag, and that there are no `<b>` tags remaining (confirming this second condition on other pages as well)


sebastien-comeau (approved):
Tested locally and it looks good!

PR #295, opened 2023-02-01
Pre-commit hook with lint-staged

sebastien-comeau:
### Description

List of proposed changes:

- Using pre-commit hook with [lint-staged](https://github.com/okonet/lint-staged)

### Additional Notes

The main motivation is to only apply lint and prettier on git staged files.
https://prettier.io/docs/en/precommit.html#option-1-lint-stagedhttpsgithubcomokonetlint-staged


krischarbonneau (approved):
Looks good

PR #299, opened 2023-02-02
Updated DoB field to use fieldset, updated htmlFor for year input

krischarbonneau:
### Description

This PR addresses the issues from the a11y audit item linked in the above ADO task. Item 3 was already done in a previous PR so this really just addresses item 1 since item 2 makes no sense (asking us to put the feedback section inside a `<p>`  would place the `<h2>` inside the `<p>` tag which is not allowed; maybe they just want the link inside? More clarity needed here).

Item 2 was to place the date of birth fields inside a `<fieldset>` and to fix the label for the year input since there was a typo.

1. Pull in branch
2. Type `npm run dev`
3. Navigate to `/status`
4.  Boot up NVDA (or another screen reader) and tab to the DoB field ensuring that you hear something like `Applicant date of birth, combo box year` when tabbing through the inputs

### Additional Notes
Currently we're using CSS to prepend required fields with `*`. If we're going to maintain this pattern then we will need to eventually scrap the CSS and manually add a span with `aria-hidden=true` where needed instead since with our current implementation, the star gets announced by screen readers, which isn't a good experience.

### DIFF
--- a/components/DateSelectField.tsx
+++ b/components/DateSelectField.tsx
@@ -178,12 +178,12 @@ const DateSelectField: FC<DateSelectFieldProps> = ({
           message={errorMessage}
         />
       )}
-      <div className="flex flex-col space-y-2 sm:flex-row sm:space-x-2 sm:space-y-0">
+      <fieldset className="flex flex-col space-y-2 sm:flex-row sm:space-x-2 sm:space-y-0">
         <DateSelect
           ariaDescribedby={getAriaDescribedby()}
           dateSelectLabelId={dateSelectLabelId}
           error={!!errorMessage}
-          id={id}
+          id={`${id}-year`}
           label={t('common:date-select-field.year')}
           onChange={handleOnDateSelectChange}
           options={yearOptions}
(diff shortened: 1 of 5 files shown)


components/DateSelectField.tsx
```diff
@@ -178,12 +178,12 @@ const DateSelectField: FC<DateSelectFieldProps> = ({
...
         />
       )}
-      <div className="flex flex-col space-y-2 sm:flex-row sm:space-x-2 sm:space-y-0">
+      <fieldset className="flex flex-col space-y-2 sm:flex-row sm:space-x-2 sm:space-y-0">
         <DateSelect
           ariaDescribedby={getAriaDescribedby()}
           dateSelectLabelId={dateSelectLabelId}
           error={!!errorMessage}
-          id={id}
+          id={`${id}-year`}
```
sebastien-comeau, line 186: This was done on pursose otherwise the error message link doesn't work. A11y shouldn't care about element's id.
`http://localhost:3000/en/status#dateOfBirth`

sebastien-comeau: If we want to keep the `-year` suffix then we need to modify how we build the error links.

sebastien-comeau: I've added the `mapping` code when building the error summry items. It fix the mentionned issue.

```tsx
getErrorSummaryItems(formikErrors, t).map((item) => {
    if (item.feildId !== 'dateOfBirth') return item
    // field id should target the year select input
    return { ...item, feildId: 'dateOfBirth-year' }
}),
```

sebastien-comeau (approved):

PR #301, opened 2023-02-03
Update asterisk in InputLabel and add aria-labelled-by to InputField

krischarbonneau:
### Description
This PR updates the `InputLabel` component to have a conditional that renders an asterisk if the input is required instead of using CSS and uses `aria-hidden="true"` to hide this from the screen reader so it doesn't get announced. I've also updated `aria-label` to `aria-labelledby` in `InputField` so the other fields get properly announced.

1. Pull in branch
2. Type `npm run dev`
3. Start up NVDA (or another screen reader)
4. Navigate to `/status` and tab through all the inputs, confirming that the labels are properly announced and the asterisk is ignored.

### Additional Notes
We should be taking notes from [this site ](https://a11y-101.com/development/required), specifically the second or third option since having both the asterisk and the word `(required)` isn't really necessary or, well, required.

### DIFF
--- a/components/InputLabel.tsx
+++ b/components/InputLabel.tsx
@@ -17,9 +17,11 @@ const InputLabel: FC<InputLabelProps> = ({
 }) => {
   return (
     <label id={id} htmlFor={htmlFor} className="mb-2 block font-bold">
-      <b className="text-accent-error mr-0.5" aria-hidden="true">
-        *
-      </b>
+      {required && (
+        <b className="text-accent-error mr-0.5" aria-hidden="true">
+          *
+        </b>
+      )}
       {label}
       {required && (
         <strong className="text-accent-error">&nbsp;{textRequired}</strong>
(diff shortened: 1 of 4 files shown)


components/InputLabel.tsx
```diff
@@ -16,11 +16,12 @@ const InputLabel: FC<InputLabelProps> = ({
...
 }) => {
   return (
-    <label
-      id={id}
-      htmlFor={htmlFor}
-      className={`mb-2 block font-bold ${required ? 'required' : ''}`}
-    >
+    <label id={id} htmlFor={htmlFor} className="mb-2 block font-bold">
+      {required && (
+        <b className="text-accent-error mr-0.5" aria-hidden="true">
```
sebastien-comeau, line 21: Change this html to this:

```tsx
<span className="text-accent-error" aria-hidden="true">
  {'* '}
</span>
```
`'* '` ensure the correct space between the asterisk and the label's text and `b` tag is not needed because the label have the bold class.

sebastien-comeau (approved):

PR #349, opened 2023-03-07
bump dependencies

sebastien-comeau:
### Description

List of proposed changes:

- bump dependencies


gregory-j-baker:
Will approve but you should know that these changes don't really do anything.

The caret matcher in `package.json` matches only the major version.. so when you do `"@tanstack/react-query": "^4.26.1"`, you're effectively doing: `"@tanstack/react-query": "^4"`.

The only reason you see anything updated in the project is because the `package-lock.json` is being updated to whatever the latest version that matches `^4` is right now.

gregory-j-baker (approved):
See comment.

PR #388, opened 2023-05-11
add data-gc-analytics-exempt link attribute

sebastien-comeau:
### Description

List of proposed changes:

- add data-gc-analytics-exempt link attribute in order to exempt outbound link clicks from analytics tracking

### DIFF
--- a/src/components/check-status-responses/CheckStatusShippingCanadaPost.tsx
+++ b/src/components/check-status-responses/CheckStatusShippingCanadaPost.tsx
@@ -34,6 +34,7 @@ export const CheckStatusShippingCanadaPost: FC<
             </p>
             <p>
               <ExternalLink
+                data-gc-analytics-exempt={true}
                 href={t('status-check-tracking.link.canada-post', {
                   trackingNumber: encodeURIComponent(trackingNumber),
                 })}
--- a/src/components/check-status-responses/CheckStatusShippingFedex.tsx
+++ b/src/components/check-status-responses/CheckStatusShippingFedex.tsx
@@ -33,6 +33,7 @@ export const CheckStatusShippingFedex: FC<CheckStatusShippingFedexProps> = ({
             </p>
             <p>
               <ExternalLink
+                data-gc-analytics-exempt={true}
                 href={t('status-check-tracking.link.fedex', {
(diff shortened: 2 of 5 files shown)


src/components/check-status-responses/CheckStatusShippingCanadaPost.tsx
```diff
@@ -34,6 +34,7 @@ export const CheckStatusShippingCanadaPost: FC<
             </p>
             <p>
               <ExternalLink
+                data-gc-analytics-exempt=""
```
gregory-j-baker, line 37: I feel like this would be a little more sense:

```suggestion
                data-gc-analytics-exempt={true}
```

src/components/check-status-responses/CheckStatusShippingFedex.tsx
```diff
@@ -33,6 +33,7 @@ export const CheckStatusShippingFedex: FC<CheckStatusShippingFedexProps> = ({
             </p>
             <p>
               <ExternalLink
+                data-gc-analytics-exempt=""
```
gregory-j-baker, line 36: See comment above.

gregory-j-baker (changes requested):

src/components/check-status-responses/CheckStatusShippingCanadaPost.tsx
sebastien-comeau: It is rendered at AA team asked for.
image

gregory-j-baker: Setting it to `true` is valid and more expressive than `""`

https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML

sebastien-comeau: Ok, will fix but I don't care either will do the job.

src/components/check-status-responses/CheckStatusShippingFedex.tsx
sebastien-comeau: Fixed for you personal pleasure. You can need eat a big bag of chips.

gregory-j-baker (approved):

PR #430, opened 2023-06-26
Feature/external alerts

stefan-oconnell:
Added an AlertBLock component that fetches alerts from principle publisher JSON file, depending on whether they are page specific or should be displayed on all pages.

Accounts for header order (H1, H2, etc.) so should not break HTML semantics.


package.json
```diff
@@ -33,6 +33,7 @@
     "react-focus-on": "^3.8.1",
     "react-i18next": "^13.0.1",
     "react-idle-timer": "^5.7.2",
+    "react-markdown": "^8.0.7",
```
sebastien-comeau, line 36: Any reason to use `react-markdown` isntead of `markdown-to-jsx`? YDG and SJ are both using `markdown-to-jsx`.
https://probablyup.com/markdown-to-jsx/

src/pages/api/alerts.ts
```diff
@@ -0,0 +1,54 @@
...
+import { Alert, AlertJsonResponse } from '../../lib/types'
+import { getLogger } from '../../logging/log-util'
+
+const logger = getLogger('get-alerts')
+
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse<Alert[] | string>
+) {
+  if (!process.env.ALERT_JSON_URI) {
```
sebastien-comeau, line 12: I'm suggesting to log a warning message and returning a 404 http status code.

gregory-j-baker: A REST collection endpoint shouldn't return 404, in my opinion. It should return an empty collection.

sebastien-comeau (approved):

PR #445, opened 2023-07-11
Tweaked alerts so pages is optional

stefan-oconnell:
Made pages optional for alerts, so full-site alerts just don't define a page list instead of adding 'all' to it which didn't feel right.

Made failed alerts return 500 with an empty list, seems to work as expected.

### DIFF
--- a/src/pages/api/alerts.ts
+++ b/src/pages/api/alerts.ts
@@ -9,35 +9,35 @@ export default async function handler(
   req: NextApiRequest,
   res: NextApiResponse<Alert[] | string>
 ) {
-  if (!process.env.ALERT_JSON_URI) {
-    logger.error('ALERT_JSON_URI must not be undefined, null or empty')
-    throw Error(
-      'process.env.ALERT_JSON_URI must not be undefined, null or empty'
-    )
-  }
-
-  if (req.method !== 'GET') {
-    res.status(405).send(`Invalid request method ${req.method}`)
-    return
-  }
-
-  const query = req.query
-  const { page } = query
-
-  let now = new Date()
-
   try {
+    if (!process.env.ALERT_JSON_URI) {
+      logger.error('ALERT_JSON_URI must not be undefined, null or empty')
+      throw Error(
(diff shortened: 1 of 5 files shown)


src/pages/api/alerts.ts
```diff
@@ -27,28 +27,35 @@ export default async function handler(
   let now = new Date()
 
   try {
```
sebastien-comeau, line 29: Move the Try at the begining (before `if (!process.env.ALERT_JSON_URI) {`)

src/pages/api/alerts.ts
```diff
@@ -27,28 +27,35 @@ export default async function handler(
   let now = new Date()
```
sebastien-comeau, line 27: Use `const` and not `let`

src/pages/api/alerts.ts
```diff
@@ -27,28 +27,35 @@ export default async function handler(
...
-        textEn: alert.textEn,
-        textFr: alert.textFr,
-        type: alert.type,
-      }))
+    let validAlerts = alertData?.jsonAlerts.filter(
+      (alert) =>
+        new Date(alert.validFrom) <= now && new Date(alert.validTo) >= now
+    )
+
+    let pageAlerts = page
```
sebastien-comeau, line 40: Use `const` and not `let`.

src/pages/api/alerts.ts
```diff
@@ -27,28 +27,35 @@ export default async function handler(
...
+      ? validAlerts.filter((alert) =>
+          alert.pages?.find((alertPage) => alertPage === page)
+        )
+      : validAlerts.filter(
+          (alert) => alert.pages === undefined || alert.pages?.length === 0
+        )
+
+    let alerts: Alert[] = pageAlerts.map((alert) => ({
```
sebastien-comeau, line 48: Use `const` and not `let`.

sebastien-comeau (changes requested):

sebastien-comeau (approved):

PR #538, opened 2023-10-24
chore(deps): bump dependencies

sebastien-comeau:
### Description

List of proposed changes:

- chore(deps): bump dependencies


package.json
```diff
@@ -43,25 +43,25 @@
...
     "autoprefixer": "^10.4.16",
     "axe-core": "^4.8.2",
     "babel-jest": "^29.7.0",
-    "cypress": "^13.3.1",
+    "cypress": "^13.3.2",
     "cypress-axe": "^1.5.0",
     "cypress-multi-reporters": "^1.6.3",
-    "eslint": "8.51.0",
-    "eslint-config-next": "^13.5.5",
+    "eslint": "8.52.0",
```
gregory-j-baker, line 57:
```suggestion
    "eslint": "^8.52.0",
```

(to match all the others)

package.json
```diff
@@ -19,20 +19,20 @@
...
-    "@tanstack/react-query": "^5.0.0",
+    "@tanstack/react-query": "^5.0.5",
     "cookies-next": "^4.0.0",
     "date-fns": "^2.30.0",
     "formik": "^2.4.5",
-    "i18next": "^23.5.1",
+    "i18next": "^23.6.0",
     "markdown-to-jsx": "^7.3.2",
-    "next": "13.5.5",
+    "next": "13.5.6",
```
gregory-j-baker, line 28:
```suggestion
    "next": "^13.5.6",
```

(to match all the others)

gregory-j-baker: This is what `package-lock.json` is for.. if the docker build isn't picking up the correct version from `package-lock.json`, then the docker build is misconfigured.

package.json
sebastien-comeau: Fixed

package.json
sebastien-comeau: We need to install `next` package only to be able to run the application in the final build. I've tested many scenario with @ken-blanchard and this is the only way we could make sure the final build had the proper next version when starting the container.

https://stackoverflow.com/a/45566871

sebastien-comeau: I've added a comment in the package.json for next version.
https://github.com/DTS-STN/passport-status/pull/538/commits/6cd5e192ef3c8091f5f406bed14635c96af25d75

blanchak: If the intention was to ONLY install next, this command does not achieve the expected results. It installs all dependencies in package.json

`npm install --no-save next `

In other projects I've reverted to use `npm ci --omit=dev`. The command `npm ci` respects the lock file.

blanchak: AFAIK There is no clean way to install a single dependency from your package.json using npm while having it respect package-lock.json.

sebastien-comeau: Will do the same as you've suggested.

sebastien-comeau: Fixed

gregory-j-baker (approved):

PR #721, opened 2024-09-09
Revert node version to 20.x

gregory-j-baker:
Reverting NodeJS version to 20.x (LTS)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
-FROM node:22.7.0-bookworm-slim AS base
+ARG BASE_IMAGE=docker.io/node:20.17.0-bookworm-slim
+FROM $BASE_IMAGE AS base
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -56,4 +57,4 @@ EXPOSE 3000
 
 ENV PORT 3000
 
-CMD npm run start
+CMD ["npm", "run", "start"]
```
sebastien-comeau, line 60: Why?

sebastien-comeau (approved):

Dockerfile
gregory-j-baker: https://testdriven.io/tips/9c233ead-5f34-4b95-b416-84f80f3ef5e7/

sebastien-comeau: TIL, thanks

PR #786, opened 2024-11-27
ADO#4864 -- Status title; Documents must have <title> element to aid in navigation

AdamP-HRSDC-RHDCC:
### Description

List of proposed changes:

- Added in dynamic loaded of status titles based on the status code received (same title as header titles)

### DIFF
--- a/src/pages/status.tsx
+++ b/src/pages/status.tsx
@@ -33,7 +33,11 @@ import InputField from '../components/InputField'
 import Layout from '../components/Layout'
 import Modal from '../components/Modal'
 import { removeCheckStatus } from '../lib/removeCheckStatus'
-import { CheckStatusApiRequestQuery } from '../lib/types'
+import {
+  CheckStatusApiRequestQuery,
+  CheckStatusApiResponse,
+  StatusCode,
+} from '../lib/types'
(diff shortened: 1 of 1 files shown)


src/pages/status.tsx
```diff
@@ -164,13 +164,32 @@ const Status = () => {
     router.push('/landing')
   }, [router])
 
+  function getTitleHeader(code: string | undefined): string {
```
ken-blanchard, line 167: This does not account for "response-no-result" scenario. Difference between checkStatusResponse being undefined or null.

image

AdamP-HRSDC-RHDCC: Updated the cases to include null

ken-blanchard: Updated version still isn't correct.

 switch (checkStatusResponse?.status) will evaluate to undefined in both null|undefined cases.
 
 Instead, early return on undefined.
 
if( checkStatusResponse === undefined) return t('header')

// else do the switch.

AdamP-HRSDC-RHDCC: Added in conditional to return for only undefined

ken-blanchard (approved):

PR #798, opened 2024-12-09
Updated text to remove file number from get status page

AdamP-HRSDC-RHDCC:
### Description

List of proposed changes:

- Updated text to remove file number

image
image


AdamP-HRSDC-RHDCC:
Fixed in another PR

PR #820, opened 2025-01-07
Stefan/phase 2 timeline

stefan-oconnell:
Created the initial version of the timeline component.

Will have to make some tweaks/implement it on the pages once we lockdown the datamodel we'll be receiving.


src/components/TimelineEntry.tsx
sebastien-comeau, line 1: Use `divide-y` to [Add borders between stacked children](https://tailwindcss.com/docs/divide-width#add-borders-between-stacked-children) instead of adding logic with `position` prop.

src/components/TimelineEntry.tsx
```diff
@@ -0,0 +1,160 @@
...
+  return position == 'last' ? 'border-transparent' : style
+}
+
+const svgStyles = {
+  done: '-translate-x-8',
+  current: '-translate-x-8',
+  future: '-translate-x-8',
+}
+
+const SVG = (type: TimelineEntryStatus, background: boolean | undefined) => {
```
sebastien-comeau, line 40: It would be a good idea to start using [Font-Awesome](https://docs.fontawesome.com/v5/web/use-with/react) icons instead of hardcoding svg in our code.

src/components/TimelineEntry.tsx
```diff
@@ -0,0 +1,160 @@
...
+          className={`${type == 'done' ? 'translate-y-6' : 'translate-y-1'} my-6 -translate-x-4`}
+        >
+          {type == 'current' || (position == 'last' && type == 'done') ? (
+            <strong>
+              <p>{step}</p>
+            </strong>
+          ) : (
+            <p>{step}</p>
+          )}
+          {type == 'done' ? <p>{date}</p> : null}
```
sebastien-comeau, line 153: Use `===` sign and [Logical AND operator (&&)](https://react.dev/learn/conditional-rendering#logical-and-operator-). Consider using the [(Date) Time element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time) over a paragraph.


```suggestion
          {type === 'done' && <p>{date}</p>}
```

sebastien-comeau (commented):
Use strict equality operator (`===`).

src/components/TimelineEntry.tsx
stefan-oconnell: In this case we don't want borders between our elements, all the positioning is just to make things line up. It's unfortunately not simple since the "easiest" way to do this timeline was using a border element with some images put on top of it:

image

stefan-oconnell: Although, looking at this again now I've added a note to the work item to try and simplify this when we go to integrate the actual feed, since it could probably be done "top down" from each timeline entry instead of "middle-out" which could help make the alignment easier.

src/components/TimelineEntry.tsx
stefan-oconnell: Won't pull this in immediately, but I agree in principle we should have a more standardized system for these SVG's. But at least if it's something we made, we can't run afoul of any usage restrictions, and it does make the dependencies simpler. I'll add an investigation item to the backlog to look into what things would look like with an icon pack, and whether business would approve the change in look.

src/components/TimelineEntry.tsx
```diff
@@ -0,0 +1,160 @@
...
+          />
+        </div>
+        <div className={`${svgStyles[type]} w-8 content-center`}>
+          {SVG(type, background)}
+        </div>
+        <div
+          className={`${type === 'done' ? 'translate-y-5' : 'translate-y-1'} my-6 -translate-x-4`}
+        >
+          {type === 'current' || (position == 'last' && type == 'done') ? (
+            <strong>
```
sebastien-comeau, line 147: `<strong>` should be inside paragraph

sebastien-comeau (changes requested):

src/components/TimelineEntry.tsx
stefan-oconnell: Done

sebastien-comeau: Forgot to push you commit?

sebastien-comeau (approved):

PR #828, opened 2025-01-24
Stefan/printing timeline updates

stefan-oconnell:
## [ADO-4677](https://dev.azure.com/DTS-STN/passport-status/_workitems/edit/4677)

Updated wording for the processing page as per work item.

Re-worked timeline logic and status pages to be closer to what the final form will be once we align with CDO.


src/components/TimelineEntry.tsx
```diff
@@ -126,6 +126,27 @@ const TimelineEntry = ({
...
+          <p className="">
+            <strong>{step}</strong>
+          </p>
+        )
+      }
+      return <p className="">{step}</p>
+    }
+
+    return <p>{step}</p>
+  }
```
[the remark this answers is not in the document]
gregory-j-baker: Also.. why not extract it into an in-file component (`<EntryStep>`) to be more idiomatic?

src/components/check-status-responses/CheckStatusPrinting.tsx
```diff
@@ -1,18 +1,77 @@
...
+              onClick={backButtonHandler}
+              text={t('status:previous')}
+              style="primary"
+            />
+          </div>
+        </div>
+        {timelineData && (
+          <div className="hidden w-full justify-center sm:flex">
+            <div className="-mt-6">
+              <Timeline entries={timelineData} />
```
gregory-j-baker, line 69: How does having `<Timeline>` here twice (also on line 32) behave in screen readers? Will a screen reader see and read both?

src/pages/status.tsx
```diff
@@ -190,6 +197,114 @@ const Status = () => {
     }
   }
 
+  // This is just a first draft of this logic.
+  // We'll ultimately have to come up with logic based
+  // on both the status + the dates (in case they get out of sync).
+  const getTimelineEntries = (
```
gregory-j-baker, line 203: I threw this function at ChatGPT and it spit out a much more concise and cleaner version. Depending on how stable this code is, that might be worth doing.

src/pages/status.tsx
```diff
@@ -190,6 +197,114 @@ const Status = () => {
...
+    }
+
+    return entries
+  }
+
+  const getStatusComponent = (response: CheckStatusApiResponse | null) => {
+    if (response !== null) {
+      let timelineData = getTimelineEntries(response)
+
+      switch (response.status) {
```
gregory-j-baker, line 274: This is more of a nitpick, so feel free to ignore..

I usually prefer wrapping each case block in curly braces for better maintainability, which helps when future changes might involve adding more logic to a block. For example:

``` typescript
switch (thing) {
  case 'Case1': {
    return doSomething();
  }

  case 'Case2': {
    throw new ApplicationError('barf!');
  }
}
```

This helps avoid subtle bugs that could be introduced when adding new code to a case block.

src/components/TimelineEntry.tsx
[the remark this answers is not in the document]
stefan-oconnell: Done. I didn't split it out because sometimes I feel like we're making TOO many components. But looking at the code, I take your point it looks kinda out of place. I did split it out, but there's probably an argument that for things we aren't re-using elsewhere, prematurely splitting them into a component makes the code harder to reason about when hopping between so many files.

src/pages/status.tsx
stefan-oconnell: True, but I pulled this from the component I deleted, so it was more or less existing code. And I don't think it'll ever be more than just a simple mapping of enum => component. So won't bother adding any bloat at this point.

src/components/check-status-responses/CheckStatusPrinting.tsx
stefan-oconnell: To my knowledge, if an element is set as hidden, screen readers will ignore it and it's children.

src/components/TimelineEntryContent.tsx
gregory-j-baker, line 1: This component can be a subcomponent inside of `TimelineEntry.tsx`, there's no reason to create a separate file for it.

gregory-j-baker (approved):

PR #884, opened 2025-04-15
Fixed the "Update information and try again" button

eric-pitre:
### Description

On the 99 status page, APPLICATION_NO_LONGER_MEETS_CRITERIA, the "Update information and try again" was returning the user to the landing page.

That button is displayed in the CheckStatusNoRecord component, which is used in two scenarios: Then there is an actual 99 code or when invalid data was entered and no status was returned. In both those case, if the user clicks that button, it should return the user to the form.

### DIFF
--- a/src/pages/status.tsx
+++ b/src/pages/status.tsx
@@ -158,15 +158,11 @@ const Status = () => {
   const handleTryAgainClick: MouseEventHandler<HTMLButtonElement> = useCallback(
     async (e) => {
       e.preventDefault()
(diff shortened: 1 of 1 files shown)


src/pages/status.tsx
```diff
@@ -155,18 +155,16 @@ const Status = () => {
     [formikErrors, t],
   )
 
+  console.log('checkStatusResponse: ', checkStatusResponse)
```
ken-blanchard, line 158: Remove.

PR #888, opened 2025-04-22
Eric/5634   passport has been mailed   text change

eric-pitre:
### Description

Added missing word in French content for passport shipped via Canada Post.


eric-pitre:
PR against wrong branch

